PR #2, opened 2021-09-20
Fix load_path so it finds digested files

brenogazzola:
Making files always digested broke load path's find. This fixes it


dhh:
When/why would you want to resolve a digested file path?

dhh:
I see the problem. Solved via 2b6599c51d743dc86d67d5e5832446cc48bc62f1.

PR #8, opened 2021-09-22
Use namespaced name for global initializer

frodsan:
It will be easier to see where the initializer comes from.


frodsan:
It seems like it also fixes some of the tests in https://github.com/rails/rails/pull/43261. I only had time to test it locally.

PR #10, opened 2021-09-29
Do not digest already digested files

brenogazzola:
@dhh Not sure how you will prefer to handle this one:

If I have an already digested file like `trix_chunk-abcdef012345678.js` in `/build`, then `Asset` will use the entire thing as the `logical_path`, but `Server` will only pass to `LoadPath` the string `trix_chunk.js`. Either `Asset` or `LoadPath` have to strip the digest too. Which do you prefer?


dhh:
I'm concerned that `-([0-9a-f]{7,128})` will pickup things like `javascript-somethingabovesevencharacters.js` and not digest. We actually have the same problem with the change made to Sprockets. I think it's good we didn't do a release there yet!

brenogazzola:
Ok, I'll submit a PR for Sprockets too.

dhh:
I guess `-[digest].digested.js` is even safer. Highly unlikely to snare files you didn't intend to catch with this.

dhh:
Yes, can you do a PR with test for the server change?

PR #11, opened 2021-10-07
Cache load path and sweep on change

brenogazzola:
Adds caching to the load path, and watches for changes in all load paths to perform a cache sweep.


lib/propshaft/load_path.rb
```diff
@@ -27,9 +28,19 @@ def manifest
     end
   end
 
+  # Returns an instance ActiveSupport::EventedFileUpdateChecker configured to clear the cache of the load_path
+  # when the directories passed during its initialization have changes. This is used in development
+  # and test to ensure the map caches are reset when javascript files are changed.
+  def cache_sweeper
```
dhh, line 34: I think we should change this to use the Rails file watcher, like importmap-rails did. It too started out using the EventedFileUpdatedChecker, but that requires the listen gem, which in turn has native dependencies. Would be better if we can do without that: https://github.com/rails/importmap-rails/blob/main/lib/importmap/map.rb#L65

brenogazzola: 🤦‍♂️ I read through importmaps commit history and simply stropped when I found the sweeper, instead of checking if you changed it later. I've updated.

Another thing: I've read through the code of the FileUpdateChecker API and it seems it does not support wildcard file extensions, so I've added all mime types that rails considers valid. Not sure how's the performance for that though. 

Is there a better way to tell which extensions to watch?

PR #13, opened 2021-10-09
Serve pre digested files and add tests

brenogazzola:
1. Ensure server can find pre digested files;
2. Add header `Content-Type` with `Vary` to ensure proper caching;
3. Add tests for server;
4. Add new method `stale?` to `asset`. Always returns `false` for predigested files.

### DIFF
--- a/lib/propshaft/asset.rb
+++ b/lib/propshaft/asset.rb
@@ -32,6 +32,10 @@ class Propshaft::Asset
     end
   end
 
+  def fresh?(digest)
+    self.digest == digest || already_digested?
+  end
+
   def ==(other_asset)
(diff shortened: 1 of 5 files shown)


lib/propshaft/asset.rb
```diff
@@ -32,6 +32,10 @@ def digested_path
     end
   end
 
+  def stale?(digest)
```
dhh, line 35: Better to flip this to fresh? so you don't need to do a negative in the one place it's used.

brenogazzola: Oh right. I remember that controllers had `stale?` and forgot about `fresh_when` when I was choosing a name.
Done.

PR #15, opened 2021-10-14
Help users migrating from Sprockets

brenogazzola:
While migrating from Sprockets and replacing image_url, asset_url, Et al., with simple urls I was constantly making mistakes and getting NoMethodError when the code called `digested_path` on nil because the asset was not found. 

So let's be a little more helpful to the developers and point which is the missing asset and the most probable solution


dhh:
Let's bring the missing error to the other places as well where we look up assets.

rmacklin:
As a follow-up to #16, I've opened #20 and included the integration test there with some more context as to why I think it's valuable (it actually originated in https://github.com/rails/importmap-rails/issues/42#issuecomment-929844040).

PR #16, opened 2021-10-23
Raise missing asset error in more places

brenogazzola:
Add missing asset error to the resolvers.

### DIFF
--- a/test/propshaft/resolver/dynamic_test.rb
+++ b/test/propshaft/resolver/dynamic_test.rb
@@ -12,7 +12,9 @@ class Propshaft::Resolver::DynamicTest < ActiveSupport::TestCase
       @resolver.resolve("one.txt")
   end
 
-  test "resolving missing asset returns nil" do
-    assert_nil @resolver.resolve("nowhere.txt")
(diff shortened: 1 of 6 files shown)


test/propshaft/resolver/dynamic_test.rb
```diff
@@ -12,7 +12,9 @@ class Propshaft::Resolver::DynamicTest < ActiveSupport::TestCase
       @resolver.resolve("one.txt")
   end
 
-  test "resolving missing asset returns nil" do
-    assert_nil @resolver.resolve("nowhere.txt")
+  test "resolving missing asset raises a custom error" do
+    assert_raise Propshaft::MissingAssetError do
+      assert_nil @resolver.resolve("nowhere.txt")
```
rmacklin, line 17: The `assert_nil` here is misleading and unnecessary after the method was changed to raise an exception:
```suggestion
      @resolver.resolve("nowhere.txt")
```
However, rather than making that change, I've proposed a small refactor in #20: https://github.com/rails/propshaft/commit/feec9cf1261d8d8efcc2e62cb223b8326d7e5659

PR #19, opened 2021-10-29
Add reveal task

brenogazzola:
Log all assets in the load paths.

I saw some `.coffee` files when I tested it and was a bit suprised. Now I'm wondering if this task should take an optional parameter to print `path` instead of `logical_path` so we can see where the asset is coming from, or if we should add another task to print all the paths from which Propshaft will pull assets from?


dhh:
Like the idea of an additional `reveal:full` that shows the physical path 👍

PR #22, opened 2021-11-07
Ignore url() inside comments

tohosaku:
propshaft now also handles "url()" in comments when compiling stylesheets.
If the asset does not exist at the specified path, even if it is in a comment, Propshaft::MissingAssetError will be raised.
It is inconvenient when commenting out some code blocks during development.

### DIFF
--- a/lib/propshaft/compilers/css_asset_urls.rb
+++ b/lib/propshaft/compilers/css_asset_urls.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+
+require 'strscan'
 require "propshaft/errors"
 
 class Propshaft::Compilers::CssAssetUrls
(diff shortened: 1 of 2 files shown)


dhh:
Hmm, this seems like a lot of code, and potential slowdown, to add just so we don't do the replacement in the CSS url. What I'd actually rather see is a way where we just don't raise on these conversions, if there's not a matching asset. Maybe we could warn instead?

cc @brenogazzola

brenogazzola:
I'm on the team that prefers that broken things scream at me to fix them instead of just warning. A warning is easy to miss in the middle of the debug level logging that happens in development, and then you keep wondering why the image is not showing up, and later why propshaft didn't adjust the url in the css.

I'll see if I can change the regex to exclude these single line comments and multi line comments without impacting performance.

brenogazzola:
@tohosaku Did you get that problem with an internal CSS or a public library? If it's the latter, can you link to it please?

dhh:
I'm using the path of warning with the source mapping conversion in #28 (as well as stripping). I think we could convert the CSS asset compiler to do the same, @brenogazzola.

brenogazzola:
Will do.

PR #26, opened 2021-11-20
Rewrite CSS source map comments to include digest

aaronjensen:
Similar to https://github.com/rails/sprockets-rails/pull/479/files

This was the only thing that prevented propshaft from being plug and play for us.

This only applies to CSS, but it could be adapted to JS if desired.

### DIFF
--- a/benchmarks/css_asset_urls
+++ b/benchmarks/css_asset_urls
@@ -23,7 +23,10 @@ file = URI.parse("https://github.com/StartBootstrap/startbootstrap-new-age/archi
 assets = ActiveSupport::OrderedOptions.new
 assets.paths     = [ dir + "/startbootstrap-new-age-master/dist" ]
 assets.prefix    = "/assets"
-assets.compilers = [ [ "text/css", Propshaft::Compilers::CssAssetUrls ] ]
+assets.compilers = [
+  [ "text/css", Propshaft::Compilers::CssAssetUrls ],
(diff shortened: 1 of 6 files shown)


dhh:
Thanks for working on this @aaronjensen! I already had a version sitting that I hadn't pushed that dealt with both JS and CSS. But appreciate you taking a swing at this too ✌️❤️

aaronjensen:
@dhh right on. I think though that the version that just got published uses `//` for CSS comments, which isn't valid. You may need to have separate compilers for JS and CSS or allow for `/*` comments.

aaronjensen:
Given that it's uncommon (rare? never?) for JS to use `/*` and impossible for css to use `//` it seems like it'd be a better design to have two classes that do exactly what they should do rather than a more complex regex with a conditional to determine what to do (when we already know to do thing 1 for JS and thing 2 for CSS and can make that decision at compiler registration time). Does that seem reasonable?

dhh:
Just seems a bit much to duplicate all of this, simply to vary on the start of the regexp. But I suppose we could just make an inheritance tree with an abstract super class and two concrete subclasses where just the REGEXP is varied.

dhh:
Just altered the regexp via https://github.com/rails/propshaft/pull/30.

aaronjensen:
The regex and the emitted string both that vary. I could capture the comment prefix and reuse it. The comment suffix will remain where it is.

ah, just saw your edit -- yep, thanks!

PR #31, opened 2021-11-22
Warn on missing asset in compiler

brenogazzola:
If an asset cannot be found when processing a CSS file, warn and replace with `url('')` instead of raising.

I did a quick test and the empty url does not seem to generate any extra network requests.


dhh:
Kinda thinking that in this case it's probably better to let the original url() go through because then you'll create a proper 404 as well. That's different from source maps because those are optional. If someone is referring to url() in css and the asset isn't there, things are broken.

brenogazzola:
Sorry, I misunderstood what you meant. Adjusted to preserve the original url.

PR #33, opened 2021-11-24
Add 'clean' rake task

brenogazzola:
Implements the `clean` rake task. Adapted from Sprockets version


brenogazzola:
Aha! Finally fixed that test.

dhh:
Do we also need to attach to assets:clean?

PR #38, opened 2021-12-02
Propshaft::Server#inspect

### LINKED ISSUE #37: bin/rails routes: Propshaft::Server emits a lot of noise
This is more of a "polish" kind of issue, but `bin/rails routes` inspects `Propshaft::Server` when it prints out the `/assets` entry:

rails-routes-propshaft-example

If `Propshaft::Server` is a singleton, perhaps the output should just be `Propshaft::Server`, which would look like this:

```
           Prefix Verb URI Pattern        Controller#Action
                       /assets            Propshaft::Server
          widgets GET  /widgets(.:format) widgets#index
```

I'd be happy to submit a PR, but I'm not sure whether implementing `#inspect` on `Propshaft::Server` is an ideal solution in this case, since `#inspect` is useful in many other scenarios besides the Rails route inspector.

Here is the relevant code in the route inspector: https://github.com/rails/propshaft/issues/37

Side note: Engines/Railties don't appear to have this problem, since the route inspector ends up calling `inspect` on the classes, rather than instances.

ntl:
The Rails route inspector (typically invoked via `bin/rails routes`)
uses `#inspect` to print the action associated with routes that mount
Rack applications. The output now looks like this:

```
     Prefix Verb URI Pattern        Controller#Action
                 /assets            Propshaft::Server
    widgets GET  /widgets(.:format) widgets#index
```

Resolves #37

### DIFF
--- a/lib/propshaft/server.rb
+++ b/lib/propshaft/server.rb
@@ -27,6 +27,10 @@ class Propshaft::Server
     end
   end
 
+  def inspect
+    self.class.inspect
+  end
+
   private
     def extract_path_and_digest(env)
       full_path = Rack::Utils.unescape(env["PATH_INFO"].to_s.sub(/^\//, ""))
--- a/test/propshaft/server_test.rb
+++ b/test/propshaft/server_test.rb
@@ -53,10 +53,6 @@ class Propshaft::ServerTest < ActiveSupport::TestCase
     assert_equal 404, last_response.status
   end
 
-  test "inspect" do
-    assert_equal "Propshaft::Server", @server.inspect
-  end
-
   private
     def default_app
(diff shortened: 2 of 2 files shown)


ntl:
I'm unsure about the value of the automated test, and I am not sure I would have included it in my own project. I included it here since it's easier for a maintainer to remove a test that exists than it is to create a test for an implementation written by a contributor.

dhh:
Fine to skip the test. Agree it doesn't offer much/anything.

dhh:
I think the diff is now deleting the whole server test?

ntl:
Ah, sorry, that was a gaffe on my part. It is corrected.

Issue #41, opened 2021-12-04
v0.4.2 appears to be broken


rmacklin:
Upgrading from propshaft v0.4.1 to v0.4.2 introduced this error on application initialization:
```
undefined local variable or method `paths' for #<Propshaft::Railtie:0x00005cd9e1dc97c8> (NameError)
```
from this line:
https://github.com/rails/propshaft/blob/b47f43b713c65399a7243e6c382fb9c454ad2c45/lib/propshaft/railtie.rb#L39

<details>

<summary>Full stack trace from trying to load a rails console</summary>

```
$ bundle exec rails c
Traceback (most recent call last):
        37: from bin/rails:4:in `<main>'
        36: from /home/richard/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:31:in `require'
        35: from /home/richard/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:22:in `require_with_bootsnap_lfi'
        34: from /home/richard/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/loaded_features_index.rb:100:in `register'
        33: from /home/richard/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `block in require_with_bootsnap_lfi'
        32: from /home/richard/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/bootsnap-1.9.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `require'
        31: from /home/richard/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/rails-14f190f33710/railties/lib/rails/commands.rb:18:in `<main>'
        30: from /home/richard/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/bundler/gems/rails-14f190f33710/railties/lib/rails/command.rb:48:in `invoke'
[...]

PR #43, opened 2021-12-04
Improve the registration of the Engines' assets path in propshaft

rafaelfranca:
I'm an not making the mistake to commit this without someone trying it.

@dhh or @rmacklin can you check if this works in the same way as before?

### DIFF
--- a/lib/propshaft/railtie.rb
+++ b/lib/propshaft/railtie.rb
@@ -1,15 +1,5 @@
 require "rails"
-require "rails/railtie"
 require "active_support/ordered_options"
-
-# FIXME: There's gotta be a better way than this hack?
-class Rails::Engine < Rails::Railtie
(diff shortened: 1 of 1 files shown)


rmacklin:
With the integration test from #20, this change [appears to pass CI](https://github.com/rmacklin/propshaft/runs/4418733714?check_suite_focus=true#step:4:11) - it doesn't break in the way that 66ce0f889377b81a28b1dd1fbe245868a047d46f broke the test suite (https://github.com/rails/propshaft/issues/41)

brenogazzola:
I've tested it against a production app already running Sprockets and the output of `rake assets:reveal` is identical, so it seems its correctly picking the assets from gems.

rafaelfranca:
Thank! @rmacklin can you open a new PR extracting the integrations tests from #20?

rmacklin:
@rafaelfranca Done: #44

PR #46, opened 2021-12-07
Test against rails v7.0.0.rc1

rmacklin:
Let's make sure our test suite is running against [the new release candidate](https://weblog.rubyonrails.org/2021/12/6/Rails-7-0-rc-1-released). Among other things, rc1 includes [the removal of the sprockets-rails dependency from rails](https://github.com/rails/rails/pull/43261), so our test app will only have propshaft installed (which better resembles real applications using propshaft).


Gemfile.lock
```diff
@@ -106,43 +107,35 @@ GEM
...
     rake (13.0.6)
     reline (0.2.7)
       io-console (~> 0.5)
-    sprockets (4.0.2)
-      concurrent-ruby (~> 1.0)
-      rack (> 1, < 3)
-    sprockets-rails (3.4.1)
-      actionpack (>= 5.2)
-      activesupport (>= 5.2)
-      sprockets (>= 3.0.0)
```
rmacklin, line 145: Bye sprockets! 👋

PR #57, opened 2022-01-09
Add support for SRI

### LINKED ISSUE #56: Support for Subresource Integrity Checks (SRI)
Similar to https://github.com/rails/sprockets-rails#sri-support
Currently adding `integrity: true` passes the option as is

brenogazzola:
Adds support for Subresource Integrity through the `integrity` option in `javascript_include_tag`
https://www.w3.org/TR/SRI

Closes #56

### DIFF
--- a/lib/propshaft/load_path.rb
+++ b/lib/propshaft/load_path.rb
@@ -23,7 +23,10 @@ class Propshaft::LoadPath
   def manifest
     Hash.new.tap do |manifest|
       assets.each do |asset|
-        manifest[asset.logical_path.to_s] = asset.digested_path.to_s
+        manifest[asset.logical_path.to_s] = {
+          "digested_path" => asset.digested_path.to_s,
+          "integrity" => asset.integrity
+        }
       end
     end
   end
--- a/lib/propshaft/resolver/dynamic.rb
+++ b/lib/propshaft/resolver/dynamic.rb
@@ -17,5 +17,13 @@ module Propshaft::Resolver
         asset.content
       end
     end
+
+    def integrity(logical_path)
+      if asset = load_path.find(logical_path)
+        asset.integrity
+      else
+        raise Propshaft::MissingAssetError.new(logical_path)
+      end
+    end
   end
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,6 +3,7 @@ PATH
   specs:
     propshaft (0.6.4)
       actionpack (>= 7.0.0)
+      actionview (>= 7.0.0)
       activesupport (>= 7.0.0)
       rack
       railties (>= 7.0.0)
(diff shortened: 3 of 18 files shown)


lib/propshaft/load_path.rb
```diff
@@ -22,7 +22,10 @@ def assets(content_types: nil)
   def manifest
     Hash.new.tap do |manifest|
       assets.each do |asset|
-        manifest[asset.logical_path.to_s] = asset.digested_path.to_s
+        manifest[asset.logical_path.to_s] = {
+          digested_path: asset.digested_path.to_s,
+          integrity: asset.integrity
+        }.stringify_keys
```
dhh, line 28: Why would the keys not already be strings if you're using #to_s on the logical path?

lib/propshaft/resolver/dynamic.rb
```diff
@@ -7,11 +7,24 @@ def initialize(load_path:, prefix:)
...
     end
+
+    def integrity(logical_path)
+      asset(logical_path) do |asset|
+        asset.integrity
+      end
+    end
+
+    private
+      def asset(logical_path)
```
dhh, line 22: I think this form a little strange. Why not just do`find_asset(logical_path).integrity`, and then have `find_asset` raise if missing?

lib/propshaft/load_path.rb
brenogazzola: Without `stringify_keys` the test below needs to use a symbol for `digested_path`. I thought looked weird having to know that the first key was a string and the second was a symbol

```ruby
test "manifest" do
    @load_path.manifest.tap do |manifest|
      assert_equal "one-f2e1ec14d6856e1958083094170ca6119c529a73.txt", manifest["one.txt"]["digested_path"]
      assert_equal "nested/three-6c2b86a0206381310375abdd9980863c2ea7b2c3.txt", manifest["nested/three.txt"]["digested_path"]
    end
  end
```

lib/propshaft/resolver/dynamic.rb
brenogazzola: Fixed. I thought it would be bad to have duplicate code like that.

lib/propshaft/load_path.rb
rafaelfranca: But isn't this hash static?

```ruby
       manifest[asset.logical_path.to_s] = {
          "digested_path" => asset.digested_path.to_s,
          "digested_path" => asset.integrity
        }
```

would wold in the same way, with no runtime cost.

brenogazzola: Fixed. My bad, I'm so used to symbols everywhere and stringify for external services that I tend to forget the fat arrow.

brenogazzola:
Closed. Check reasoning at #56

PR #59, opened 2022-01-11
Do not mount the server if public_file_server is not enabled

aaronjensen:
There wasn't an obvious place to test this, so I omitted the test.

### DIFF
--- a/lib/propshaft/railtie.rb
+++ b/lib/propshaft/railtie.rb
@@ -28,8 +28,10 @@ module Propshaft
 
       app.assets = Propshaft::Assembly.new(app.config.assets)
 
-      app.routes.prepend do
-        mount app.assets.server => app.assets.config.prefix
+      if app.config.public_file_server.enabled
+        app.routes.prepend do
+          mount app.assets.server => app.assets.config.prefix
+        end
       end
 
       ActiveSupport.on_load(:action_view) do


rafaelfranca:
This will break the library since the public server is never enabled in development or tests. This is using the wrong config. You actually wants the propshaft server disabled when using the public server, not enabled. I’m going to revert.

aaronjensen:
@rafaelfranca I'm not sure I follow, what am I missing? Rails enables it by default in test:

```
  # Configure public file server for tests with Cache-Control for performance.
  config.public_file_server.enabled = true
  config.public_file_server.headers = {
    "Cache-Control" => "public, max-age=#{1.hour.to_i}"
  }
```

That's in `test.rb` of a newly generated app.

rafaelfranca:
True. `public_file_server` controls the `ActionDispatch::Static` server, which you probably want enable in production even though you don't want the Propshaft server there, for example in a heroku setup. We should be using a different option. spockets users `config.assets.compile`. Since Propshaft doesn't compile, I think we should use a `config.assets.server`

aaronjensen:
@rafaelfranca Got it, I forgot that propshaft gets out of the way upon precompile so its server is orthogonal. Thanks for setting us straight. I opened #61

PR #61, opened 2022-01-11
Asset server is disabled in production

aaronjensen:


aaronjensen:
@rafaelfranca is this what you had in mind?

rafaelfranca:
Yes! Thank you

aaronjensen:
Great, thank you.

lib/propshaft/railtie.rb
```diff
@@ -12,6 +12,7 @@ class Railtie < ::Rails::Railtie
       [ "text/javascript", Propshaft::Compilers::SourceMappingUrls ]
     ]
     config.assets.sweep_cache = Rails.env.development?
+    config.assets.server = !Rails.env.production?
```
dhh, line 15: Think we need the opposite here. Turning the server on in dev + test, so it'll remain off in other envs like staging and beta.

aaronjensen: https://github.com/rails/propshaft/pull/62

PR #73, opened 2022-02-19
Resolve already digested assets using their undigested path

printezisn:
This is especially useful when assets are already being digested by a module bundler (webpack, etc.), but we want to resolve them based on their base name.

### DIFF
--- a/lib/propshaft/asset.rb
+++ b/lib/propshaft/asset.rb
@@ -4,6 +4,8 @@ require "action_dispatch/http/mime_type"
 class Propshaft::Asset
   attr_reader :path, :logical_path, :version
 
+  DIGESTED_NAME_PATTERN = /-([0-9a-zA-Z]{7,128})\.digested/
+
   def initialize(path, logical_path:, version: nil)
     @path, @logical_path, @version = path, Pathname.new(logical_path), version
   end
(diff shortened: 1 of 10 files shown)


brenogazzola:
Doesn’t the feature to skip digest already solve that problem?

printezisn:
@brenogazzola You mean the fact that when a file has `-[digest].digested` in its name, then it's not digested again, right ? Unfortunately, in this case we can't resolve the asset path by using its base name.

For example, let's say that we have `already_digested-50f7d266afaecd96b52d41cbfaa49d03444bd845.digested.js` which is an already digested asset. We can't reference it like `javascript_include_tag 'already_digested'`, because it's not found in the manifest.

It would be really useful if propshaft could do the above, so that we could still use it in scenarios where we have to let webpack handle the asset building (along with digesting) and want to resolve the asset paths with propshaft.

brenogazzola:
Yeah, there is this limitation. The entryfiles are not supposed to come pre digested, only the chunks. Both webpack and esbuild have the necessary configuration options for that. 

The problem with resolving files based on their undigested names is that the assets:clean task keeps older versions of each file (up to 3). This is done to keep compatibility with sprockets based deployment setups (eg: Heroku) and avoid 404 errors during the transition period. 

By indexing them in the hash without their digest, you will end up with a single version, instead of all versions. Worse: the version that is kept will be random, not the latest one.

printezisn:
I see! I didn't think about that case. Thank you very much for explaining this to me!
I'll close the PR then and I'll try to find a different solution to my problem.

PR #35, opened 2021-11-28
Add upgrade guide for propshaft

brenogazzola:
Tasks:

- [x] Intro
- [x] Migrate from `webpacker` to `jsbundling-rails`
- [x] Migrate from `sass-rails` to `cssbundling-rails`
- [x] Migrate from `sprockets` to `propshaft`
- [x] Tidying up

Issue #76, opened 2022-02-23
Upgrade documentation typo on Section 3


stevegeek:
Hi! I was just following the [sprockets upgrade guide](https://github.com/rails/propshaft/blob/main/UPGRADING.md) and spotted a typo:

On Section 3, step 1, it says `Replace 'sass-rails' with 'propshaft';` - guess this should be `Replace 'sprockets-rails' with 'propshaft';`?

Also, in case its useful for future docs: 

while this guide is clearly about Sprockets -> propshaft, I was moving from a `webpacker` with no `sprockets-rails` setup, in which case if you first remove `webpacker` (Steps 1/2 on Section 1) it is not possible to run the `jsbundling-rails` install script on Step 3 as Rails will fail to boot with `NoMethodError: undefined method 'assets' for #<Rails::Application::Configuration` if the environment contains any `assets` config. So need to do the install before removing `webpacker` or add the sprockets railtie temporarily.

Just finished migrating a large app to `js/cssbundling-rails` & `propshaft` 🎉  Thanks!

brenogazzola:
> On Section 3, step 1, it says `Replace 'sass-rails' with 'propshaft';` - guess this should be `Replace 'sprockets-rails' with 'propshaft';`?

Nope. It's really `sass-rails`. An app running the standard Sprockets for CSS, Webpacker for JS will have `sass-rails` in its Gemfile, not `sprockets-rails`.

> While this guide is clearly about Sprockets -> propshaft, I was moving from a webpacker with no sprockets-rails setup

We could use another upgrade guide, for someone who is fully on Webpacker, instead of Sprockets/Webpacker. Would you mind opening a PR for that? I think you could duplicate the current and remove things you didn't need, update the ones that were different, and add the ones that were not mentioned.

stevegeek:
Thanks @brenogazzola , I see. OK sure, I have my notes from when I did the migration, I will flesh it out into a guide for review. In the mean time should I close this issue?

PR #79, opened 2022-03-02
Instruct removing Sprockets from Gemfile

sedubois:
I suppose the upgrade guide introduced in #35 should instruct to remove Sprockets from the Gemfile?

Section 3 point 4 instructs to remove `app/assets/config/manifest.js`, which yields `Sprockets::Railtie::ManifestNeededError` when Sprockets is still present.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -132,7 +132,7 @@ Sass might raise deprecation warnings depending on what features you are using (
 
 Start by following these steps:
 
-1. Remove `sprockets` and `sass-rails` from the gemfile and add `propshaft`;
+1. Remove `sprockets`, `sprockets-rails`, and `sass-rails` from the gemfile and add `propshaft`;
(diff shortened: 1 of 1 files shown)


brenogazzola:
Might be better to mention `sprockets-rails` too since https://github.com/rails/propshaft/issues/76 had a similar issue with it. I assumed a sprockets/webpacker project (which the guide is targeted at) would have `sass-rails` instead of any of the sprockets gems, but with more people raising the same point...

sedubois:
Yes, the only reason sprockets became an explicit dependency in our project was to temporarily pin its version before later allowing an upgrade to version 4.

stevegeek:
@sedubois Yeah something similar happened to me (just opened #81 with some other updates). Btw one suggestion here is maybe to change `gemfile` to `Gemfile` since I think it is mostly written capitalised.

brenogazzola (approved):

PR #83, opened 2022-03-07
Identify relative protocol asset URLs

barrywoolgar:
Hello. Propshaft does not appear to be detecting URLs using the double-slash relative protocol, eg:

```
@font-face { font-family: Inter; src:url("//fonts.gstatic.com/s/inter/abc.eot") } # abridged
```

This leads them to be processed as local assets, with the subsequent warning:

```
Unable to resolve '//fonts.gstatic.com/s/inter/abc.eot' for missing asset '/fonts.gstatic.com/s/inter/s/inter/abc.eot' in application.css
```

This PR adds "//" to the URL regex's negative lookahead.

Thanks!

### DIFF
--- a/lib/propshaft/compilers/css_asset_urls.rb
+++ b/lib/propshaft/compilers/css_asset_urls.rb
@@ -3,7 +3,7 @@
 class Propshaft::Compilers::CssAssetUrls
   attr_reader :assembly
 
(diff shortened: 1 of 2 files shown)


brenogazzola:
I was not aware this was a [valid url format](https://developer.mozilla.org/en-US/docs/Web/CSS/url()). What would the GET url look like when the browser makes the request?

brenogazzola:
Since Propshaft is handling `url` directly, I believe it should support everything that `url` supports, even if it's considered anti-pattern.

Thank you for the fix and for checking Propshaft!

Issue #90, opened 2022-04-29
change is not detected in @import css file 


equivalent:
If I understood correctly the idea of Portshaft  (pls correct me if I'm wrong ) then the way how to include css files would be:

in `app/assets/application.css`

```
/* Application styles */
@import url('/bootstrap.css');
@import url('/my-custom.css');
```

in `app/assets/bootstrap.css` I have the  content of `bootstrap.css` from [Bootstrap 5 download page](https://getbootstrap.com/docs/5.0/getting-started/download/) 

in `app/assets/my-custom.css`:

```
p {
  color: red;
}
```

my `app/views/layout/application.css`

```erb
<!DOCTYPE html>
<html>
  <head>
    <title>Myapp</title>
    <meta name="viewport" content="width=device-width,initial-scale=1">
    <%= csrf_meta_tags %>
    <%= csp_meta_tag %>

    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
    <%= javascript_importmap_tags %>
  </head>

  <body>
    <h1>Propshaft test</h1>

    <div class="text-primary">
      If this text is blue, Bootstrap5 css works via Portshaft
    </div>

    <p>If this text is in color then @include css custom file works</p>
  </body>
</html>
```

So far so good, I run rails server and load the localhost:3000  

<img width="439" alt="Screenshot 2022-04-29 at 15 22 55" src="https://user-images.githubusercontent.com/721990/165952818-e67ff037-bcf2-403f-853d-e8d9ad150eee.png">

you can see the `<p>` is in red

<img width="766" alt="Screenshot 2022-04-29 at 15 24 26" src="https://user-images.githubusercontent.com/721990/165953068-b578ed72-7524-4dfc-95bf-6ba15f93d56d.png">

Now I change the `app/assets/my-custom.css`:

```
p {
  color: yellow;
}
```

 **I save the file and reload my localhost:3000**

<img width="408" alt="Screenshot 2022-04-29 at 15 25 26" src="https://user-images.githubusercontent.com/721990/165953237-c3c682bc-84a4-4088-9567-7f03e1c64498.png">

<img width="756" alt="Screenshot 2022-04-29 at 15 26 01" src="https://user-images.githubusercontent.com/721990/165953349-5137d1f5-c623-4829-845b-5b96d6be9d74.png">

As you can see the change that I want the collor of `<p>` to be yellot was not picked up. 

[...]

PR #94, opened 2022-05-04
README update - avoid isses related to browser cache during development

equivalent:
related to hours spent figuring out  https://github.com/rails/propshaft/issues/90 :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,16 @@ On the other hand, if you're already bundling JavaScript and CSS through a Node-
 
 But for greenfield apps using the default import-map approach, Propshaft can also work well, if you're able to deal with vanilla CSS.
 
+## Avoid issues related to browser cache
+
+Propshaft digests your file based on the content and your browser will by default cache the digested file.
+There are cases where if a CSS file references
+to another file via `@import url('/other-file.css)`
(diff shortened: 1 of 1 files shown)


dhh:
@brenogazzola We should think about a better solution here. But one does not immediately come to mind!

brenogazzola:
Since the server is only used in development we could change the cache-control header to no cache or no store for css files. I’ll test it and see what happens.

brenogazzola:
Possible PR to fix this: #96

brenogazzola:
I've been testing this our outside of rails, and it seems this is a normal problem with the `@import` rule. Browsers cache the files and don't notice if you change the `@import` inside them. I thought they might skip caching or something like that, but no. So we can't cheat and copy their behavior. 

All solutions I've tried so far get stuck in one of two:
1 - You need stable fingerprints in production (no cache busting on every deploy)
2 - You might have multiple levels of `@import` or cyclical `@import`

dhh:
Could we simply send no-cache headers in development?

equivalent:
Yes in fact the no-cache header does the trick ! 🎉. Creating new PR with solution => https://github.com/rails/propshaft/pull/124

closing this one

PR #98, opened 2022-06-22
Handle SVG mask URL with encoded anchor

thibaudgg:
This patch ensures that CSS assets with inlined SVG containing a mask
URL with an HTML encoded anchor are properly handled by the
`CssAssetUrls` compiler.

This avoids printing warning messages like:
```
Unable to resolve '%23c' for missing asset '%23c' in XYZ.css
```

### DIFF
--- a/lib/propshaft/compilers/css_asset_urls.rb
+++ b/lib/propshaft/compilers/css_asset_urls.rb
@@ -3,7 +3,7 @@
 class Propshaft::Compilers::CssAssetUrls
   attr_reader :assembly
 
-  ASSET_URL_PATTERN = /url\(\s*["']?(?!(?:\#|data|http|\/\/))([^"'\s?#)]+)([#?][^"')]+)?\s*["']?\)/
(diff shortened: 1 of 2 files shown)


brenogazzola:
Hello @thibaudgg, thanks for checking Propshaft. Two questions:

1 - The test for this compiler should use only CSS rules and attributes, since those are the files that the compiler is applicable to, but you seem to have used an HTML tag. Can you show a longer example of how this would be used in an CSS file?

2 - `%23c` seems like an odd thing to hard code in the regex. Is this like `data` that is part of an specification and will always be that value?

brenogazzola:
Thanks, that makes it easier to understand. 

Let's shorten your example (`<svg mask='url(#myMask)'></svg>`) and then try to keep this test as similar to the others as possible. Not sure how that was encoded so I'm doing this by hand, but it would be something like this?
```css
compiled = compile_asset_with_content(%({ background: url("data:image/svg+xml;charset=utf-8,%3Csvg mask='url(%23cMyMask)'%3E%3C/svg%3E"); }))
```

thibaudgg:
Yeah, your example is better. I updated the PR and went with just `%23MyMask` (without the `c` as it was just the name of the mask previously).

thibaudgg:
@brenogazzola thank you for reviewing and merging my PR so quickly! 👍🏻

Issue #99, opened 2022-07-08
Proposal for capturing digested assets path


rainerborene:
Hey there! I would like to share some bits of code that we implemented at our company and it could be merged into the official `propshaft` gem. Right now there is no way to reference digested assets from webpack, rollup.js or any other build system. On the other side, `propshaft` already does recognize digested assets with `.digested.` pattern on the filename. The motivation for this feature is to be able to use `preload_link_tag` or reference any kind of asset generated from external builders. No more talk, here is the code.

```ruby
# lib/propshaft/digested.rb
module Propshaft
  CACHE = {}

  module Digested
    refine Propshaft::LoadPath do
      def resolve_digested(asset_name)
        Propshaft.resolve_digested_from(asset_name, source: assets_by_path)
      end
    end

    refine Propshaft::Resolver::Static do
      def resolve_digested(asset_name)
        Propshaft.resolve_digested_from(asset_name, source: parsed_manifest)
      end
    end

    refine Propshaft::Resolver::Dynamic do
      def resolve_digested(asset_name)
        load_path.resolve_digested(asset_name)
      end
    end
  end

  def self.resolve_digested_from(asset_name, source:)
    CACHE[asset_name] ||= source.find do |logical_path, _|
      filename, extname = asset_name.split(".")
      pattern = /\A#{filename}-([0-9a-zA-Z]{7,128})\.digested\.#{extname}\z/
      pattern.match?(logical_path)
    end&.first
  end
end

# app/helpers/propshaft_helper.rb
module PropshaftHelper
  using Propshaft::Digested

  def digested_path(path)
    Rails.application.assets.resolver.resolve_digested(path) || raise(Propshaft::MissingAssetError.new(path))
  end
end
```

Let me know if this sounds good so I can start working on a pull request. :D

dhh:
Yes, I like this. cc @brenogazzola

brenogazzola:
Sure! @rainerborene Could you also post some sample code of this in action?

From what I'm seeing, the `resolve_digested_from` will allow us to pass `logo.png` and have propshaft find `logo-DIGEST-digested.png`. I'm a bit confused about the `digested` though. If the builder already generated the file in that format, shouldn't it have already compiled the JS files to replace a reference to `logo.png` with `logo-DIGEST.digested.png`? This is what happens when we configure webpack/esbuild to generate entryfiles and chunkfiles with the `.digested` pattern.

rainerborene:
[...]

Issue #48, opened 2021-12-10
Propshaft ESM Modules chunks not resolved correctly (works with sprockets-rails)


navidemad:
**application.js**:
<img width="385" alt="Screenshot 2021-12-10 at 16 41 36" src="https://user-images.githubusercontent.com/6686865/145601264-deb9e36d-f4b9-40fa-a3d7-af5197287fe5.png">

**app/views/layouts/application.html.erb**:
<img width="513" alt="Screenshot 2021-12-10 at 16 46 20" src="https://user-images.githubusercontent.com/6686865/145601555-894c9506-23df-492f-ae5f-dba1e64c0342.png">

**Google Chrome console / networks tab**
<img width="963" alt="Screenshot 2021-12-10 at 16 42 55" src="https://user-images.githubusercontent.com/6686865/145601260-0441871d-35c6-4545-983e-cde5f69458be.png">
<img width="728" alt="Screenshot 2021-12-10 at 16 42 45" src="https://user-images.githubusercontent.com/6686865/145601263-c4d7e6b6-554c-484a-af47-094d087c9b4f.png">

**Scenarios**:

Gemfile: Propshaft uncommented, sprockets-rails commented
`ESBUILD_MODULES=1 ./bin/dev`
Output: 🔴

Gemfile: Propshaft commented, sprockets-rails uncommented
`ESBUILD_MODULES=1 ./bin/dev`
Output: 🟢

Gemfile: Propshaft commented, sprockets-rails uncommented
`./bin/dev`
Output: 🟢

Gemfile: Propshaft uncommented, sprockets-rails commented
`./bin/dev`
Output: 🟢

Minimalist application to reproduce the issue: https://github.com/navidemad/PropshaftApp

brenogazzola:
If you precompile your assets you will probably see that sprockets breaks too, because it only adds digests to the file names after precompilation. 

I think we are going to need a jscompiler like we have for csscompiler to solve this.

thanosbellos:
Hi

By changing the regex in [asset.rb](https://github.com/rails/propshaft/blob/main/lib/propshaft/asset.rb) to match the one in the this [sprockets pull request](https://github.com/rails/sprockets/pull/726)

and using the following esbuild configuration from the sample repo above:

```javascript
"use strict";

import esbuild from "esbuild";
import path from "path";
import glob from "glob";

const __dirname = path.resolve();

var entryPoints = glob
  .sync(path.resolve(__dirname, `app/javascript/*.js`))
[...]

PR #100, opened 2022-07-26
feat(paths): allow assets to be found using their non digested path

rainerborene:
After talking to @brenogazzola this morning I decided to implement a more convenient solution. This is a different take to solve the issue #99. Instead of adding a new helper like `propshaft_digested_path` we are normalizing the `manifest` artifact and allowing any assets to be found using their non digested path. This works across all Rails helpers like a charm. This means that you can preload and link chunks and assets generated from webpack and other builders. 

Say you have a chunk named `trumbowyg-ccbb42265c65b6f9a38e.digested.js` stored on your `app/assets/builds` folder now you can reference it using only the filename, for example, `preload_link_tag "trumbowyg.js"`. And the digested path is also stored on the `manifest` file so that you can serve it through the Rails server or for backwards compatibility. :)

Pretty neat, huh?

### DIFF
--- a/lib/propshaft/asset.rb
+++ b/lib/propshaft/asset.rb
@@ -2,10 +2,15 @@ require "digest/sha1"
 require "action_dispatch/http/mime_type"
 
 class Propshaft::Asset
+  PREDIGESTED_REGEX = /-([0-9a-zA-Z]{7,128}\.digested)/
+
   attr_reader :path, :logical_path, :version
 
   def initialize(path, logical_path:, version: nil)
-    @path, @logical_path, @version = path, Pathname.new(logical_path), version
+    @path         = path
+    @digest       = logical_path.to_s[PREDIGESTED_REGEX, 1]
+    @logical_path = Pathname.new(@digest ? logical_path.sub("-#{@digest}", "") : logical_path)
+    @version      = version
   end
 
   def content
@@ -25,23 +30,14 @@ class Propshaft::Asset
   end
 
   def digested_path
-    if already_digested?
-      logical_path
-    else
-      logical_path.sub(/\.(\w+)$/) { |ext| "-#{digest}#{ext}" }
-    end
+    logical_path.sub(/\.(\w+)$/) { |ext| "-#{digest}#{ext}" }
   end
 
   def fresh?(digest)
-    self.digest == digest || already_digested?
+    self.digest == digest
   end
 
   def ==(other_asset)
     logical_path.hash == other_asset.logical_path.hash
   end
-
-  private
-    def already_digested?
-      logical_path.to_s =~ /-([0-9a-zA-Z]{7,128})\.digested/
-    end
 end
--- a/lib/propshaft/load_path.rb
+++ b/lib/propshaft/load_path.rb
@@ -48,7 +48,8 @@ class Propshaft::LoadPath
         paths.each do |path|
           without_dotfiles(all_files_from_tree(path)).each do |file|
             logical_path = file.relative_path_from(path)
-            mapped[logical_path.to_s] ||= Propshaft::Asset.new(file, logical_path: logical_path, version: version)
+            asset = Propshaft::Asset.new(file, logical_path: logical_path, version: version)
+            mapped[asset.logical_path.to_s] ||= asset
           end if path.exist?
         end
       end
--- a/lib/propshaft/output_path.rb
+++ b/lib/propshaft/output_path.rb
@@ -42,7 +42,7 @@ class Propshaft::OutputPath
       modified_at = [ 0, Time.now - mtime ].max
       modified_at < expires_at || limit < count
     end
-  
+
     def remove(path)
       FileUtils.rm(@path.join(path))
       Propshaft.logger.info "Removed #{path}"
--- a/lib/propshaft/server.rb
+++ b/lib/propshaft/server.rb
@@ -11,8 +11,8 @@ class Propshaft::Server
     if (asset = @assembly.load_path.find(path)) && asset.fresh?(digest)
       compiled_content = @assembly.compilers.compile(asset)
 
-      [ 
-        200, 
+      [
+        200,
         {
           "Content-Length"  => compiled_content.length.to_s,
           "Content-Type"    => asset.content_type.to_s,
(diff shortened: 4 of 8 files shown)


brenogazzola:
Currently Propshaft treats digests it generates differently from digests that the builder generates, when it comes the `logical_path`.

Generated by propshaft:
```
logical_path  => application.js
digested_path => application-abcdef123456.js
```

Generated by a builder:
```
logical_path  => application-123456abcdef.digested.js
digested_path => application-123456abcdef.digested.js
```

The hypothesis is that by making Propshaft use the same logical path in both cases, we can solve multiple problems that currently exist when integrating with builders.

@rainerborene I'm a bit crunched for time but as soon as I can I will give this a try in my own production app and then review the PR.

lib/propshaft/load_path.rb
```diff
@@ -23,7 +23,8 @@ def assets(content_types: nil)
   def manifest
     Hash.new.tap do |manifest|
       assets.each do |asset|
-        manifest[asset.logical_path.to_s] = asset.digested_path.to_s
+        manifest[asset.logical_path.to_s] ||= asset.digested_path.to_s
+        manifest[asset.non_digested_path.to_s] ||= asset.digested_path.to_s if asset.already_digested?
```
brenogazzola, line 27: I'm not really a fan of this duble indexing. I think we need to come up with a better solution for this one.

brenogazzola:
@rainerborene I've been going back and forth with this PR and I'm wondering if we can't considerably simplify everything by separating the predigest from the logical path when initializing the asset.

asset.rb
```ruby
PREDIGESTED_REGEX = /-([0-9a-zA-Z]{7,128}\.digested)\..+\z/

def initialize(path, logical_path:, version: nil)
  @path         = path
  @digest       = full_path[PREDIGESTED_REGEX, 1]
  @logical_path = digest ? logical_path.sub("-#{digest}", "") : logical_path
  @version      = version
end
```

server.rb
```ruby
def extract_path_and_digest(env)
  full_path = Rack::Utils.unescape(env["PATH_INFO"].to_s.sub(/^\//, ""))
  digest    = full_path[PREDIGESTED_REGEX, 1]
  path      = digest ? full_path.sub("-#{digest}", "") : full_path

  [ path, digest ]
end
```

This way the logical path would never have a digest in it, which means we can eliminate the existing `already_digested?` which I added a while back and we won't need this PT's undigested path.

What would happen is that a file like `application.css` would be given a digest as normal, and a file like `application-ABCDEF1.digested.js` would take on `ABCDEF1.digested` as the `@digest` instance variable.

brenogazzola:
@rainerborene Actually, I'm wondering if we shouldn't move the `extract_path_and_digest` that the `server.rb` to somewhere other classes can use.

This way server.rb, asset initializer and resolvers can all use it to get the actual logical path. I think this would also solve the double indexing in the resolver.

rainerborene:
@brenogazzola Thanks for the feedback! I have implemented the suggested improvements. Let me know if it needs further changes.

lib/propshaft/asset.rb
```diff
@@ -2,10 +2,16 @@
...
+
   attr_reader :path, :logical_path, :version
 
   def initialize(path, logical_path:, version: nil)
-    @path, @logical_path, @version = path, Pathname.new(logical_path), version
+    @path = path
+    @digest = logical_path.to_s[PREDIGESTED_REGEX, 1]
+    @digested_path = Pathname.new(logical_path) if @digest
+    @logical_path = Pathname.new(@digest ? logical_path.sub("-#{@digest}", "") : logical_path)
+    @version = version
```
brenogazzola, line 14: Now that we are normalizing the logical path to never have a digest in it, I believe that we can entirely discard the `.digested` suffix? It's already done its job in telling Propshaft that there is a digest in the file name. This would make non-digested and pre-digested logical paths look the same, and I think it would also incentivize the rest of the code to treat them the same.

By the same token, I think we don't need the `@digested_path` instance variable. Removing it (and having the logical path always `sub` instead of using a ternary operator) means extra `subs` that could have been avoided like you did here, but it also makes the code a bit tidier. At this point where we are still experimenting, I think that would be worth more than the small optimization and make it easier for others since the behavior is always the same.

```ruby
@logical_path =  logical_path.sub("-#{@digest}.digested", "")
```

lib/propshaft/server.rb
```diff
@@ -34,8 +34,9 @@ def inspect
   private
     def extract_path_and_digest(env)
       full_path = Rack::Utils.unescape(env["PATH_INFO"].to_s.sub(/^\//, ""))
-      digest    = full_path[/-([0-9a-zA-Z]{7,128})\.(?!digested)[^.]+\z/, 1]
-      path      = digest ? full_path.sub("-#{digest}", "") : full_path
+      digest = full_path[/-([0-9a-zA-Z]{7,128})\.(?!digested)[^.]+\z/, 1]
+      digest ||= full_path[Propshaft::Asset::PREDIGESTED_REGEX, 1]
+      path = digest ? full_path.sub("-#{digest}", "") : full_path
```
brenogazzola, line 39: Same as I mentioned in `asset.rb`: we can probably drop the `.digested` suffix here.

brenogazzola:
@rainerborene Looking better. Now that I look at it, seems there's a couple of things we can improve further, like completely dropping the `.digested` suffix and (for now) focusing on simpler, more consistent behavior, instead of extra optimization to avoid `.subs`.

lib/propshaft/output_path.rb
```diff
@@ -42,7 +42,7 @@ def fresh_version_within_limit(mtime, count, expires_at:, limit:)
       modified_at = [ 0, Time.now - mtime ].max
       modified_at < expires_at || limit < count
     end
-  
+
```
brenogazzola, line 45: ✂️

lib/propshaft/server.rb
```diff
@@ -11,8 +11,8 @@ def call(env)
     if (asset = @assembly.load_path.find(path)) && asset.fresh?(digest)
       compiled_content = @assembly.compilers.compile(asset)
 
-      [ 
-        200, 
+      [
+        200,
```
brenogazzola, line 15: ✂️

brenogazzola (changes requested):
Hello @rainerborene. Sorry it took me so long to get back. I've done a few extras checks and this looks good. Can you just remove a few extra spaces that were added to keep things clean, and squash again?

I'll merge when you do.

brenogazzola:
Actually, nevermind. Just noticed you removed the spaces, not added.

brenogazzola (approved):

rainerborene:
:tada:

weilandia:
This commit breaks chunking with esbuild when implemented as suggested here: https://github.com/rails/propshaft/issues/48

The reason is likely do the the fact that esbuild creates static chunks that do not have names, i.e. `chunk-123`, `chunk-456`, etc.

This is discussed here: https://github.com/evanw/esbuild/issues/1716

Issue #97, opened 2022-05-25
How do you switch Propshaft back to dynamic mode in Development?


ethan-dowler:
The docs say:
> **Precompilation in development**
> 
> Propshaft uses a dynamic assets resolver in development mode. However, when you run assets:precompile locally Propshaft will then switch to a static assets resolver. Therefore, changes to assets **will not be observed anymore** and you will have to precompile the assets each time changes are made. This is different to Sprockets.

My question is:
How do I switch it back? It seems crazy that if you run `rails assets:precompile` one time you're screwed... 🙁

brenogazzola:
Just delete the `public/assets` folder. Or modify your `bin/dev` file do add this line before the `foreman` command:
```
rm public/assets/.manifest.json
```

abevoelker:
I ran into this today as well - I couldn't figure out why my browser was seeing stale assets. Traced it back to my `public/assets` directory, then ultimately realized I must've ran `assets:precompile` just while playing around with Propshaft. Running `rails assets:clobber` is another way to "fix" the problem as the whole `public/assets` dir gets deleted.

Perhaps when precompiling assets when `Rails.env.development?` a warning message should be displayed that dynamic changes to assets will no longer be rendered until the manifest file is deleted?

brenogazzola:
@abevoelker Yup, adding a warning there is on the list of things to be done. I wouldn't reject a PR for it 😉

PR #102, opened 2022-08-03
Warn when precompiling assets in development

abevoelker:
Per #97 this adds a warning when doing `rails assets:precompile` in development

I don't know what the preferred wording is, and also the default logging in a fresh Rails 7 app seems to be writing this to `log/development.log` but not emitting it to the console's STDERR - not sure if that is desirable?

### DIFF
--- a/lib/propshaft/railties/assets.rake
+++ b/lib/propshaft/railties/assets.rake
@@ -2,6 +2,10 @@ namespace :assets do
   desc "Compile all the assets from config.assets.paths"
   task precompile: :environment do
     Rails.application.assets.processor.process
+    if Rails.env.development?
+      puts "Warning: You are precompiling assets in development. Rails will not " \
+        "serve any changed assets until you delete public/assets/.manifest.json"
+    end
   end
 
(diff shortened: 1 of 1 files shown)


brenogazzola:
Thanks @abevoelker. 

Since the message explicitly uses the word "precompiling" and "serving assets" I'm not sure placing it in the compiler is the best choice. If we have something other the `assets:precompile` using that method later, the message would not make much sense.

Let's follow the example of the `reveal` feature and move the message to the task itself and use a standard `p`. This would ensure it's printed to the terminal and always visible to the developer, even if he chose to have his logs somewhere else.

abevoelker:
@brenogazzola Thanks, that makes sense. Let me know if the force-pushed change is more like what you were thinking of. Should it be a standard `puts` like the `reveal` feature which writes to stdout, or perhaps `warn` or `$stderr.puts` that go to stderr since it's a warning? Or perhaps it doesn't matter. At a glance I see a mix in the main Rails code base.

brenogazzola (approved):

PR #107, opened 2022-08-28
Define `Rails::Application.assets=` if it doesn’t exist

dmcge:
Propshaft’s railtie sets an `Assembly` instance on `Rails::Application` via `app.assets=`. This mirrors Sprockets setting its own `Environment` instance there. But, as of https://github.com/rails/rails/pull/45878, the attr_accessor is no longer defined on the base class. So we have to define it here.

(The alternative to this is that we revert the rails PR 🤷‍♂️)


brenogazzola:
Thank you for the early warning. That PR change was reverted.

Issue #104, opened 2022-08-18
Rails asset_host config is not respected by CssAssetUrls compiler


tonydewan:
If the [asset_host config](https://guides.rubyonrails.org/configuring.html#config-asset-host) is set, the URLs generated by Propshaft within CSS files do not contain the asset host, and instead are always relative. Perhaps this is intentional, but I can't find documentation or discussion of that.

markstanley-nps:
This looks to be related to my issue #103 with the [relative_root_url](https://guides.rubyonrails.org/configuring.html#deploy-to-a-subdirectory-relative-url-root).

There is a set of logic in the [rails asset helper](https://github.com/rails/rails/blob/04972d9b9ef60796dc8f0917817b5392d61fcf09/actionview/lib/action_view/helpers/asset_url_helper.rb) that isn't part of the [css_asset_urls.rb](https://github.com/rails/propshaft/blob/467f2a13cfc3d30a3366408c8f98c2deb4efc741/lib/propshaft/compilers/css_asset_urls.rb) or [source_mapping_urls.rb](https://github.com/rails/propshaft/blob/467f2a13cfc3d30a3366408c8f98c2deb4efc741/lib/propshaft/compilers/source_mapping_urls.rb).

The compilers could defer back to the rails routine but that coupling seems wrong.

brenogazzola:
Yeah, this is currently not supported. I'll add it to our to-do list, but I'd also welcome a PR if any of you is willing to takle

tonydewan:
> Yeah, this is currently not supported. I'll add it to our to-do list...

Awesome. thanks!

> .. but I'd also welcome a PR if any of you is willing to takle

Cool! I might be able to take a swing at it.

markstanley-nps:
Thanks both
@tonydewan let me know if I can assist.

juliacollins:
@brenogazzola Are there plans for a near-term release including PR #118? I'm interested from the relative_url_root point of view.

brenogazzola:
Hey @juliacollins. PR #118 is still untested because none of us using Propshaft in production needs this feature, so I've been waiting for input from people who do need it to see if it's actually working. 

If you could test this (not necessarily in production, but somewhere that does rely on asset host) and report back if it's working or not I can move forward with this.

juliacollins:
[...]

PR #118, opened 2022-11-14
Add support for asset host

brenogazzola:
Closes https://github.com/rails/propshaft/issues/104

I'll leave this unmerged for now so others can test and give suggestions

### DIFF
--- a/lib/propshaft/compilers/css_asset_urls.rb
+++ b/lib/propshaft/compilers/css_asset_urls.rb
@@ -1,12 +1,13 @@
 # frozen_string_literal: true
 
 class Propshaft::Compilers::CssAssetUrls
-  attr_reader :assembly
+  attr_reader :assembly, :url_prefix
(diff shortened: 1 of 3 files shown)


brenogazzola:
This was requested for users whose assets are not in the app's domain, but other domains, like S3. 

This is usually used by apps that perform rolling releases and have a lot of servers. For example, if you change `application.js` and a user refreshes a page, the HTML request might go to server `web1` which has been updated and will point to `application_new.js`, but the JS request might go to server `web13` which has not been updated yet (because the rolling release is still on server `web8`) and therefore does not have that file, only `application_old.js`. This will cause a 404. Less of a problem with images, large problem with `application.js` in an SPA.

brenogazzola:
If you configure your app's `asset_host` to point to your CDN and upload the new assets to there before you start a rolling release, the problem is solved.

PR #124, opened 2022-12-29
ability to specify cache-control header in config

equivalent:
Ability to tell what cache-control header we want the assets loaded with

This will solve the issue https://github.com/rails/propshaft/issues/90

Solution proposed by @dhh  in [issue 94](https://github.com/rails/propshaft/pull/94#issuecomment-1355086241) :

Screenshot 2022-12-29 at 21 07 56

two things here:

#### 1. no-cache header

 `no-cache` works but maybe `no-store`  is more what we want ([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control))
Screenshot 2022-12-29 at 21 09 50

... I don't mind either 

#### 2. this could be simpler?

I'm not sure if I didn't overkill this. Maybe something simpler would be enough eg:

```
# lib/propshaft/server.rb
       {
          #...
          "cache-control"   => Rails.development? ? 'no-store' : "public, max-age=31536000, immutable"
       }
```

...let me know I can simplify it

### DIFF
--- a/lib/propshaft/railtie.rb
+++ b/lib/propshaft/railtie.rb
@@ -15,6 +15,11 @@ module Propshaft
       [ "text/css", Propshaft::Compilers::SourceMappingUrls ],
       [ "text/javascript", Propshaft::Compilers::SourceMappingUrls ]
     ]
+    if Rails.env.development?
+      config.assets.cache_control_header = "no-store"
+    else
+      config.assets.cache_control_header = "public, max-age=31536000, immutable"
+    end
     config.assets.sweep_cache = Rails.env.development?
     config.assets.server = Rails.env.development? || Rails.env.test?
 
--- a/lib/propshaft/server.rb
+++ b/lib/propshaft/server.rb
@@ -18,7 +18,7 @@ class Propshaft::Server
           "content-type"    => asset.content_type.to_s,
           "accept-encoding" => "vary",
           "etag"            => asset.digest,
-          "cache-control"   => "public, max-age=31536000, immutable"
+          "cache-control"   => Rails.application.config.assets.cache_control_header
         },
         [ compiled_content ]
       ]


equivalent:
🤔 wait a minute, now I'm re-reading @brenogazzola [comment](https://github.com/rails/propshaft/pull/94#issuecomment-1134633497) 

Screenshot 2022-12-29 at 21 27 37



...hmm I'm not sure about the prod but this solved the dev isuue 🤔

brenogazzola:
Video looks a bit strange to me. The digest of `application.css` did not change, which means the `@include "my-custom.css"` should still be pointing to `8ffbe1`. So how did the browser know it had to fetch `f4b376` after refresh?

equivalent:
Sorry for the late reply @brenogazzola , I'm in  middle of Job search, those damn tech interview keep me busy :)

Ok took me a while to realise what's happening:

* yes application-xxxxxxxx.css digest don't change
* yes I agree theoretically it should (as digest is calculated from content and `@import url('/my-custom.css');` is translated to `@import url('/my-custom-mmmmmmm.css');` = application.css content changed so it should be  application-yyyyyy.css ... **but it's not!**)


To see what I mean please watch https://youtu.be/cqQHyJj7few

[video showing digest dont change](https://youtu.be/cqQHyJj7few)

equivalent:
one more thing - during testing the above I've stumbled upon an edegecase: 

My PR proposes the no-store header by default for Development env(therefor no such issue as #90) 👍.

 If a developer switch the non-cache header to any form of cache header and then switch back to no-store again, his/her browser will be stuck at old cache.

To see this in action pls watch [https://youtu.be/3nmCubctHuk](https://youtu.be/3nmCubctHuk)

[video showing how browser needs cache clear when switch from cahce header to no-store](https://youtu.be/3nmCubctHuk)


Given on cache header browser will cache the old application-xxxxx.css 
When I switch to no-store header browser still cache the old application-xxxxx.css 
Therefore it will not pick up the fact the my-custom.css has a new digest

conclusion:
* My PR works  👍
* My PR solves #90 issue in development (tat means first time users will not have that issue as Development env fixed by default) 👍
* If someone tries to test if cache works, they need to clear the browser cache or do a change in application.css (so new digest appears for application-yyyy.css) 😕

simi:
:muscle: I can confirm this fixes the problem. And yes, I had to wipe the cache locally first to get this change effective.

lib/propshaft/railtie.rb
```diff
@@ -15,6 +15,11 @@ class Railtie < ::Rails::Railtie
       [ "text/css", Propshaft::Compilers::SourceMappingUrls ],
       [ "text/javascript", Propshaft::Compilers::SourceMappingUrls ]
     ]
+    if Rails.env.development?
+      config.assets.cache_control_header = "no-store"
+    else
+      config.assets.cache_control_header = "public, max-age=31536000, immutable"
```
simi, line 21: Does it make sense to have this value out-of-sync with `config.assets.sweep_cache`?
Wouldn't be easier to keep it in sync by default like

```ruby
config.assets.cache_control_header = config.assets.sweep_cache ? "no-store" : "public, max-age=31536000, immutable"
```

:question:

equivalent: yep  `sweep_cache` makes more sense 👍, I will change it later this week 😇

simi: Clear, this could be fixed by setting original value of sweep_cache to `nil` and only when it is not assigned to false explicitly, change `cache_control_header` and also self assign to `false`. Anything in here works for me. Even two independent config options as originally proposed. With sane default values per environment it would be rarely tweaked by users anyway.

equivalent: Sorry I'm getting back after 2 weeks (I had some personal issues to solve) 

Thank you for the recommendation how to fix this with `nil` `sweep_cache`. I understand 

>  Even two independent config options as originally proposed

Unless you really  want the sweep_cache nil  solution (as you proposed) I would vote for the original "two independent config options" solution.  I get it no-one will really change the values.  But personally purely to have simple straight forward  config. 

please let me know if I should do the `nil` change proposed or not

simi: As mentioned no huge preference from my side. Let's grab some maintainers attention somehow, they can decide on their own. :pray:

@brenogazzola @dhh friendly ping

brenogazzola:
> how come the digest of application-xxxxxxxx.css didn't change ? 🤷‍♂️ my guess is that the digest of application.css is calculated on a file content before content gets digested (before `my-custom.css` turns to `my-custom-nnnnnn.css`

I'm a bit worried about this part. Yes, you are right, parent digest is calculated on the original content, not the one that was changed with the digest of the children imports.

This means this solution will give you a false impression that everything is working because in the development you are seeing the changes. But when you push to production, anyone who had already fetched the parent file (eg: your CDN which is sending the files to your users) will not fetch again it again, since its digest did not change and they think it's the same file.

This means that if you are using new classes that were added to the children files, your site is now broken (at least visually).

The only feasible solution I see is one that ensures that parent digest changes when children digest change

dhh:
We need to fix this at the root. The no-cache setup isn't going to work for production either. Need to invalidate files when their dependencies have changed. There are several open approaches for that, so exploring those.

Issue #103, opened 2022-08-16
Propshaft appears to be ignoring Rails config.relative_url_root


markstanley-nps:
I have just switched to use propshaft from sprockets and it has all gone well except for one last issue.

When running in a local development environment, with the rails config.relative_url_root set, the compilers do not prepend this value onto the generated urls.

url([asset]) becomes url([config.assets.prefix]/[digested-asset]) not url([config.relative_url_root]/[config.assets.prefix]/[digested-file]) 

This means the assets cannot be found as they are under the url root.

I believe in production I can overcome this by setting the config.assets.prefix to be <config.relative_url_root>/assets as this will generate the correct path, and then the assets are served up separately.

However, the above solution does not work in development (or test) as while the correct url is generated in the css, the asset cannot be served by the development server (they are under /url_root/url_root/assets)

Is there a way of setting up the configuration so that I can use relative_url_root in development or should the compilers be including the relative_url_root in the generated path

brenogazzola:
Same as #104. We haven't tested Propshaft with `relative_url_root` yet, so it's probably not supported. I've added it to our to-do list, but would welcome a PR for it.

PR #130, opened 2023-03-06
Adds support for relative url root

markstanley-nps:
Closes https://github.com/rails/propshaft/issues/103

This builds on the changes in PR https://github.com/rails/propshaft/pull/118 to add support for relative_url_root 

It also extends both fixes to apply to the source mapping compiler. In doing so I have moved the derivation of the prefix to the Assembly to dry up the solution.

It is my first PR so let me know of anything I can do to improve it.

### DIFF
--- a/lib/propshaft/assembly.rb
+++ b/lib/propshaft/assembly.rb
@@ -8,10 +8,11 @@ require "propshaft/compilers/css_asset_urls"
 require "propshaft/compilers/source_mapping_urls"
 
 class Propshaft::Assembly
-  attr_reader :config
+  attr_reader :config, :url_prefix
 
   def initialize(config)
     @config = config
+    @url_prefix = File.join(config.relative_url_root.to_s, config.prefix.to_s).chomp("/")
   end
 
   def load_path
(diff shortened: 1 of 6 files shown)


brenogazzola:
Hello @markstanley-nps I've reverted #118 as I had misunderstood the original issue and that change was actually not needed. Setting `asset_host` already worked in Propshaft. Unfortunately this caused conflicts with your PR. Would you mind rebasing so I can check it?

markstanley-nps:
Hi @brenogazzola I have rebased the changes, the commit history isn't as clean as I would like, let me know if you want me to tidy it up.  It may be cleaner to go back and do another pull request.

brenogazzola:
Hello @markstanley-nps. Sorry it took me so long to check this.

It seems something is wrong with git history, as "Files changed" tab still shows code from #118 that is no longer on main (like the changes in `CssAssetUrls`.

Another thing I've noticed, is this line in `railtie.rb`:
```
mount app.assets.server => app.assets.config.prefix
```

Wouldn't it also need to take into consideration the relative_url_root?

markstanley-nps:
@brenogazzola no problem

It looks like it needs a rebase to resolve the conflicts so I will have a go at that and try and get a tidier history, not sure what I broke. If I can't tidy this pull request up, I will open another one.

> ```
> mount app.assets.server => app.assets.config.prefix
> ```
> 
> Wouldn't it also need to take into consideration the relative_url_root?

Not in the way we deploy and configure. We use the rack config (config.ru) to run rails in the subdirectory as below so the above is not impacted
```
map '/subdirectory' do
  run Rails.application
end
```

markstanley-nps:
Closed in favour of PR #138

PR #132, opened 2023-03-07
Prevent damage to /* ... */ comments with missing source maps

pond:
If a source map is missing in a JS or CSS file that uses the `/* ... */` style of comments, Propshaft would inadvertently substitute nothing via `gsub` for the opening comment as well as the source map part, leaving a broken trailing `*/`.

Fixed, with test coverage to prove it.


pond:
Bump on this.

pond:
Another bump here. To be clear, **Propshaft is corrupting files** without this fix, albeit in a way that is relatively benign since it breaks the comment at the very end of the file.

brenogazzola:
Thank you for the contribution @pond. Sorry it took me so long to check and merge this.

pond:
Awesome. Thanks for merging!

PR #136, opened 2023-03-27
fix: Vary HTTP header

dunglas:
`Accept-Encoding` is a request header. The intent here was to set the `Vary` header to prevent sending responses compressed with modern algorithms to old browsers and sending responses compressed with old suboptimal algorithms to modern browsers.

### DIFF
--- a/lib/propshaft/server.rb
+++ b/lib/propshaft/server.rb
@@ -16,7 +16,7 @@ class Propshaft::Server
         {
           "content-length"  => compiled_content.length.to_s,
           "content-type"    => asset.content_type.to_s,
-          "vary" => "Accept-Encoding",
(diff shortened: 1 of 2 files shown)


lib/propshaft/server.rb
```diff
@@ -16,7 +16,7 @@ def call(env)
         {
           "content-length"  => compiled_content.length.to_s,
           "content-type"    => asset.content_type.to_s,
-          "accept-encoding" => "vary",
+          "vary" => "Accept-Encoding",
```
guillaumebriday, line 19: Please keep the arrow aligned with other ones 👍

brenogazzola:
@dunglas Thanks for noticing that mistake.

PR #138, opened 2023-05-26
Adds support for relative url root

### LINKED ISSUE #103: Propshaft appears to be ignoring Rails config.relative_url_root

markstanley-nps:
Closes #103 

This replaces PR #130 to add support for relative_url_root

@brenogazzola hopefully this is easier to review.

You had commented that the below may need to account for the relative_url_root

`mount app.assets.server => app.assets.config.prefix
`

We use rack config (config.ru) to run rails in the subdirectory as below, so we don't need the above changed
```
map '/subdirectory' do
  run Rails.application
end
```
We are currently using this change in our development pipeline, and it has been deployed to production. For production the static assets are served up by a separate server under the relative url.

### DIFF
--- a/benchmarks/css_asset_urls
+++ b/benchmarks/css_asset_urls
@@ -9,7 +9,7 @@ require "open-uri"
 require_relative "./trackrod"
 require_relative "../lib/propshaft"
 require_relative "../lib/propshaft/compilers"
-require_relative "../lib/propshaft/compilers/css_asset_urls"
+require_relative "../lib/propshaft/compiler/css_asset_urls"
 
 trackrod = Trackrod.new(Dir.mktmpdir)
 trackrod.build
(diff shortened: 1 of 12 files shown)


brenogazzola (changes requested):

brenogazzola:
Thank you for your contribution!

Issue #139, opened 2023-05-29
Newly added files that are already digested aren't available in development


joker-777:
Hi, we have a couple of assets that we don't want propshaft to digest and therefore we add the pre`digested` suffix to its file name. This worked until the new version 0.7.0. [This pull request](https://github.com/rails/propshaft/pull/100) introduced changes that broke this feature.  It doesn't add new assets with the `digested` suffix the `load_path` (only the first found asset). 
The problem is that [in this line](https://github.com/rails/propshaft/pull/100/files#diff-76c66ff09264f5f7a9fdf0e703dbbc4ae3315acc3de5c7bc7607247efe774d48R52) `asset.logical_path` is always `filename.js` and not `filename-xyz.digested.js`. This is because in [this line](https://github.com/rails/propshaft/pull/100/files#diff-cea8115669d1b6f7ed0b417671e6b929c07d4f41b4e009b82182221f38b46955R12) it removes the predigested part from the logical path.

brenogazzola:
I've been considering reverting that one. Quick question: how are those files being created? Are they esbuild chunks?

joker-777:
No, they are created with webpack. If I find some time later I can try to find a fix and create a PR for it.

joker-777:
@brenogazzola I created [a little PR](https://github.com/rails/propshaft/pull/141). I didn't test it yet though.

joker-777:
@brenogazzola Just fixed the specs of my little PR.

PR #141, opened 2023-05-30
Fixes: Newly added files that are already digested aren't available in development

joker-777:
* makes sure that we don't remove the digest from the logical path if the asset is already digested
* related issue https://github.com/rails/propshaft/issues/139


joker-777:
The tests are now green.

dhh:
Was sorted in https://github.com/rails/propshaft/pull/144

PR #133, opened 2023-03-07
Source maps do not have to be on their own line

pond:
> (**NB: This PR builds on https://github.com/rails/propshaft/pull/132**. I can't see a way to do a pull request on top of the prior pull request, so the diff on this one will make more sense once https://github.com/rails/propshaft/pull/132 is completed. I had to build on it since the source map removal logic of https://github.com/rails/propshaft/pull/132 is now extended further.)

While some older resources claim that source maps must be on their own line at the end of a file and while Propshaft's Regexp for this asserts such, this is not the case. Consider:

```
sass ./app/assets/stylesheets/application.scss ./app/assets/builds/application.css --no-source-map --style compressed --load-path=node_modules
```

...in `package.json`. This results in `application.css` including a source map comment in `/* ... */` form at the *end of* the last line of minified CSS:

```
<very-long-line>...vertical-align:middle}/*# sourceMappingURL=application.css.map */
```

...noting that a terminating line feed is also present afterwards. Since this is a CSS comment appearing at the end of a line it is syntactically valid and browsers such as Chrome *do* read the source map. Under e.g. Propshaft 0.7.0, CSS source maps promptly break because it doesn't consider the not-on-own-line map comment to be valid.

* It seems that so long as the comment is the last thing in the file, whether on its own line or not, then it is processed
* Source map comments in the middle of the file are, as before, *not* processed
* The regexp is therefore updated to check for end-of-stream (`\Z`) rather than start-of-line (`^`)
* This means it might optionally capture a closing `*/` in comments of that form, so the internal private methods now need to be given both an opening and closing string for the comment captures
* Test coverage for JS with `//#` and CSS with `/* ... */` style comments included

PR #142, opened 2023-06-11
Fix sourceMappingURL in compressed/minified files

rbague:
As of right now, the `SourceMappingUrls` compiler does not work with [sass compressed](https://sass-lang.com/documentation/cli/dart-sass#style) CSS files, as the regex ...
https://github.com/rails/propshaft/blob/26082d0b8a5106a79c7e0bbb65ef780cc47bb837/lib/propshaft/compiler/source_mapping_urls.rb#L6
... used to detect sourceMappingURLS expects the comment to be in a new line (which is not the case in the compressed mode, as it outputs everything in a [single line](https://github.com/sass/dart-sass/blob/aa59a5f4a91f7dafb15c9b88b0970de7aca19db1/lib/src/executable/compile_stylesheet.dart#L195-L196)).

Following what is stated in the [Source Map Proposal](https://sourcemaps.info/spec.html), that the annotation must be _at the end of the source_, I've updated the regex so instead of looking for the comment at the start of a line, it looks for it at the end of the source, and included tests for compressed JS and CSS files.

The new regex should cover all the possible scenarios of how source map annotations are being appended to the generated CSS/JS file:
- [esbuild](https://github.com/evanw/esbuild/blob/394941be8c17494938abc05e52fb6f52c5cda6d9/internal/linker/linker.go#L611-L629)
- [PostCSS](https://github.com/postcss/postcss/blob/15c5f61aa6f72df51f78dbda724b044f616be4fe/lib/map-generator.js#L172)
- [Sass](https://github.com/sass/dart-sass/blob/aa59a5f4a91f7dafb15c9b88b0970de7aca19db1/lib/src/executable/compile_stylesheet.dart#L195-L196)
  - Even in [legacy mode](https://github.com/sass/dart-sass/blob/aa59a5f4a91f7dafb15c9b88b0970de7aca19db1/lib/src/node/legacy.dart#L401)

Let me know if there is anything I need to update to get this merged.

### DIFF
--- a/lib/propshaft/compiler/source_mapping_urls.rb
+++ b/lib/propshaft/compiler/source_mapping_urls.rb
@@ -3,10 +3,10 @@
 require "propshaft/compiler"
 
 class Propshaft::Compiler::SourceMappingUrls < Propshaft::Compiler
-  SOURCE_MAPPING_PATTERN = %r{^(//|/\*)# sourceMappingURL=(.+\.map)}
+  SOURCE_MAPPING_PATTERN = %r{(//|/\*)# sourceMappingURL=(.+\.map)((?:\s\*/)?(?:\r?\n)*)\Z}
 
   def compile(logical_path, input)
-    input.gsub(SOURCE_MAPPING_PATTERN) { source_mapping_url(asset_path($2, logical_path), $1) }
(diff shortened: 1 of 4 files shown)


aaronjensen:
Is this a duplicate of #133?

rbague:
Closing this now that #133 is merged and solves the same problem

PR #144, opened 2023-07-01
Revert "feat(paths): allow assets to be found using their non digested path (#100)"

### LINKED ISSUE #117: digested source maps
recently stumbled into this project while attempting a rails 7 update. I am using the predigested ".digested" method because my js files need to reference themselves, and am having 2 issues with the source maps.

1. The `sourceMappingURL` is being striped away: https://github.com/rails/propshaft/blob/390381548b125e8721c8aef1b9d894b7cc8bd868/lib/propshaft/compilers/source_mapping_urls.rb#L29 (quick and dirty solution in dev mode is to call `config.assets.compilers.pop` to remove `Propshaft::Compilers::SourceMappingUrls`)
2. Also the source maps will not be served without tweaking `extract_path_and_digest` - https://github.com/firien/propshaft/compare/3903815...ed67507

### LINKED ISSUE #139: Newly added files that are already digested aren't available in development

### LINKED ISSUE #143: Allow digested files with the same name prefix
ESBuild `splitting` feature generates multiple files with the same name and different hash:

```
esbuild app/typescript/src/*.* --bundle --splitting --chunk-names='chunks/[name]-[hash].digested' --format=esm ...
```
```
chunk-KJQZHWRU.digested.js
chunk-KJQZHWRU.digested.js.map
chunk-VBQABUUW.digested.js
chunk-VBQABUUW.digested.js.map
chunk-XON4CK2V.digested.js
chunk-XON4CK2V.digested.js.map
```

Propshaft will copy only one js and one map file to the `public` directory. After spending long time I figured out that it is due to the same file prefix. Had to workaround with this:

```
esbuild app/typescript/src/*.* --bundle --splitting --chunk-names='chunks/[hash]-[hash].digested' --format=esm ...
```
```
EANSPK5S-EANSPK5S.digested.js 
EANSPK5S-EANSPK5S.digested.js.map
MZ6DBTZJ-MZ6DBTZJ.digested.js
MZ6DBTZJ-MZ6DBTZJ.digested.js.map
RHSFM4N7-RHSFM4N7.digested.js
RHSFM4N7-RHSFM4N7.digested.js.map
```

I am unable to think of the reason why there should be this undocumented limitation.

brenogazzola:
Closes #143 
Closes #139 
Might close #117 
Replaces #141

While this was a nice idea, once again issues show that trying to extract digests generated by bundlers causes compatibility problems. Let's revert so esbuild chunking works again and think of another way to do this.

@joker-777 Your PR seems to revert most of the changes here, so I think it might be a better idea to revert the entire thing. I've added you as a co-author.

I'm keeping this PR open for a while before merging in case anyone involved in the listed issues/PRs wants to provide further input.

### DIFF
--- a/lib/propshaft/asset.rb
+++ b/lib/propshaft/asset.rb
@@ -2,15 +2,10 @@ require "digest/sha1"
 require "action_dispatch/http/mime_type"
 
 class Propshaft::Asset
-  PREDIGESTED_REGEX = /-([0-9a-zA-Z]{7,128}\.digested)/
-
   attr_reader :path, :logical_path, :version
 
   def initialize(path, logical_path:, version: nil)
-    @path         = path
-    @digest       = logical_path.to_s[PREDIGESTED_REGEX, 1]
-    @logical_path = Pathname.new(@digest ? logical_path.sub("-#{@digest}", "") : logical_path)
-    @version      = version
+    @path, @logical_path, @version = path, Pathname.new(logical_path), version
   end
 
   def content
@@ -30,14 +25,23 @@ class Propshaft::Asset
   end
 
   def digested_path
-    logical_path.sub(/\.(\w+)$/) { |ext| "-#{digest}#{ext}" }
+    if already_digested?
+      logical_path
+    else
+      logical_path.sub(/\.(\w+)$/) { |ext| "-#{digest}#{ext}" }
+    end
   end
 
   def fresh?(digest)
-    self.digest == digest
+    self.digest == digest || already_digested?
   end
 
   def ==(other_asset)
     logical_path.hash == other_asset.logical_path.hash
   end
+
+  private
+    def already_digested?
+      logical_path.to_s =~ /-([0-9a-zA-Z]{7,128})\.digested/
+    end
 end
--- a/lib/propshaft/load_path.rb
+++ b/lib/propshaft/load_path.rb
@@ -48,8 +48,7 @@ class Propshaft::LoadPath
         paths.each do |path|
           without_dotfiles(all_files_from_tree(path)).each do |file|
             logical_path = file.relative_path_from(path)
-            asset = Propshaft::Asset.new(file, logical_path: logical_path, version: version)
-            mapped[asset.logical_path.to_s] ||= asset
+            mapped[logical_path.to_s] ||= Propshaft::Asset.new(file, logical_path: logical_path, version: version)
           end if path.exist?
         end
       end
--- a/lib/propshaft/output_path.rb
+++ b/lib/propshaft/output_path.rb
@@ -42,7 +42,7 @@ class Propshaft::OutputPath
       modified_at = [ 0, Time.now - mtime ].max
       modified_at < expires_at || limit < count
     end
-
+  
     def remove(path)
       FileUtils.rm(@path.join(path))
       Propshaft.logger.info "Removed #{path}"
--- a/lib/propshaft/server.rb
+++ b/lib/propshaft/server.rb
@@ -11,8 +11,8 @@ class Propshaft::Server
     if (asset = @assembly.load_path.find(path)) && asset.fresh?(digest)
       compiled_content = @assembly.compilers.compile(asset)
 
-      [
-        200,
+      [ 
+        200, 
         {
           "content-length"  => compiled_content.length.to_s,
           "content-type"    => asset.content_type.to_s,
@@ -34,7 +34,7 @@ class Propshaft::Server
   private
     def extract_path_and_digest(env)
       full_path = Rack::Utils.unescape(env["PATH_INFO"].to_s.sub(/^\//, ""))
-      digest    = full_path[/-([0-9a-zA-Z]{7,128}(?:\.digested)?)\.[^.]+\z/, 1]
+      digest    = full_path[/-([0-9a-zA-Z]{7,128})\.(?!digested)[^.]+\z/, 1]
       path      = digest ? full_path.sub("-#{digest}", "") : full_path
 
       [ path, digest ]
--- a/test/fixtures/assets/first_path/file-already-abcdefVWXYZ0123456789.digested.debug.css
+++ b/test/fixtures/assets/first_path/file-already-abcdefVWXYZ0123456789.digested.debug.css
@@ -1,0 +1,5 @@
+/* this is css */
+
+.btn {
+    background-image: asset-path("archive.svg");
+}
--- a/test/propshaft/asset_test.rb
+++ b/test/propshaft/asset_test.rb
@@ -25,8 +25,7 @@ class Propshaft::AssetTest < ActiveSupport::TestCase
     assert find_asset("one.txt").fresh?("f2e1ec14d6856e1958083094170ca6119c529a73")
     assert_not find_asset("one.txt").fresh?("e206c34fe404c8e2f25d60dd8303f61c02b8d381")
 
-    assert find_asset("file-already-abcdefVWXYZ0123456789.digested.css").fresh?("abcdefVWXYZ0123456789.digested")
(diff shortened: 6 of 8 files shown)


joker-777:
@brenogazzola Thanks, then let's revert it.

midnight-wonderer:
Wow, what a coincidence. I am debugging the same issue at the moment.
Thanks, @joker-777

brenogazzola:
With no further input from anyone, and having run this in production without problems for a while, I'm merging it

aaronjensen:
What is the migration path for people that relied on this behavior?

brenogazzola:
Hey @aaronjensen can you explain how you were using it so I can see if I can think of anything? The original contributor was using it for preloads for example.

brenogazzola:
I'm not sure I understand what you are doing. You are saying instead of using propshaft's precompile to generate the manifest file you are having Vite do it?

aaronjensen:
Uhm, kind of. We have something that works. Propshaft can’t control the digest on vite/rollup assets because of how they are loaded lazily from JavaScript without a rather sophisticated propshaft compiler

aaronjensen:
> But I want to point out that everyone here use some kind of javascript tooling. That's why we choose propshaft.

No, on the contrary. Propshaft and import maps allow you to eschew JavaScript tooling, which is an appropriate response given how complex it has become. 

When we are forced to use JavaScript tooling, we are forced to deal with the things we have to deal with. My only point was "this pain is to be expected."

> For me, propshaft can't control digest of assets generated by javascript build tools sound about right as it should be.

🤷 If you are using esbuild to generate a single file, propshaft can digest things just fine. It's only necessary to have vite/webpack/whatever emit digests when the emitted JavaScript code refers to the files by name. If they never do that, you can let Propshaft do the digesting.

> I have zero issue in my use case, apart from the bugs solved by this PR.

Glad to hear it. This PR allowed us to get rid of the silly double digesting we needed to do for our async chunks that Vite generated (`chunk-abcdef-abcdef.digested.js`) but it made it so we could no longer rely on Propshaft to find the digested version of the thing we were trying to load, which was the thing the original PR that was reverted was attempting to do AFAICT. We relied on that behavior. It was removed. We had to do work to stop relying on it. 

When I asked if there was a migration path, it was to see if any thought had been given to the people that were going to be impacted by rolling something that people may have relied on. It sounds like not, it was just done, so we figured it out.

brenogazzola:
@aaronjensen I apologize for the rushed revert. I assumed the feature was a complete failure and the only use case I knew of (preloads) was a minor thing so removing the feature would not be a problem. 

We are in a bit of a frustrating place because the idea was good, but esbuild name collision is making things complicated. Each bundler behaves a different way and coming up with a way that works for all of them AND all use cases is not obvious (see the assets:clean problem)

aaronjensen:
Thanks. I don't know if I am familiar w/ the particular assets:clean problem you are referring to (though I am familiar w/ the general problem of knowing which ones to clean once they are digested... this goes way back).

One of the things that would have made our work easier today was if there was a sanctioned way to add to Propshaft's manifest. It would be relatively straightforward to translate Vite's manifest into something that Propshaft could use I'm guessing. I just wasn't sure where to start and I didn't feel like coupling to Propshaft internals (especially given this recent break).

In case it's useful to anyone, here is how we solved our problem today:

```rb
module JavascriptEntrypointHelper
  def javascript_entrypoint_tag(name)
    name = "src/#{name}.tsx"

    path = Manifest.file_path(name)

    src = path_to_javascript(path)

    tag.script(src:, crossorigin: "anonymous", type: "module", defer: true)
  end

  # This manifest is produced by Vite, which adds the digest to assets. This
  # is necessary because files import other files and we don't have a
  # Propshaft compiler that replaces those references with Propshaft digested
  # references. We attempted to have Vite emit entrypoints that did not have
  # digests, but some of the code-split chunks refer to entrypoints, so we end
  # up with the same problem. Instead, we let Vite digest all of its assets
  # and then use its manifest to produce an asset path that Rails can use.
  # - Aaron, Kelly, Tue Oct 31 2023
  module Manifest
    def self.get
      @manifest ||= load
    end

    def self.file_path(name)
      manifest = get
      entry = manifest.fetch(name)
      entry.fetch("file")
    end

    def self.load
      text = File.read(path)
      JSON.parse(text)
    end

    def self.path
      Engine.root.
        join("app/assets/builds/manifest.json")
    end
  end
end
```

midnight-wonderer:
I see now that our context is wildly different.  
Our setup heavily uses code-splitting and dynamic imports with bells and whistles, not just compiling to a single file.

I am not sure if it applies to Vite or your situation, but the way we did it is this:

* We don't emit `*.digested.*` on the entry files but everything else. Like this:
~~~javascript
{
  output: {
    filename: '[name].js',
    chunkFilename: '[id]-[contenthash:20].digested.js',
    sourceMapFilename: '[name]-[contenthash:20].digested.map',
    path: outputPath,
  },
}
~~~
The same principle also applies to CSS/SCSS.
* We emit manifest from Webpack, only for the entrypoint that is
~~~javascript
new AssetsPlugin({
  filename: 'index.json',
  removeFullPathAutoPrefix: true,
  path: outputPath,
  prettyPrint: true,
  entrypoints: true,
}),
~~~
* The manifest looks similar to this
~~~json
{
  "entry1": {
    "css": "entry1.css",
    "js": "entry1.js"
  },
  "entry2": {
    "css": "entry2.css",
    "js": "entry2.js"
  },
  "entry3": {
    "css": "entry3.css",
    "js": "entry3.js"
  }
}
~~~
Note that there is no `*.digest.*` in their names.
* We do have entry-loading helpers implemented (similar to what you are doing just now) by reading the manifest file.
~~~ruby
  def asset_pack_tags(entry_name, js_options: {}, css_options: {})
    asset_index = AssetIndexCache.instance.cached
    entry = asset_index.fetch(entry_name, {})
    [
      *entry.fetch(:css, []).then do |files|
        Array(files)
      end.map do |file|
        stylesheet_link_tag file.delete_suffix('.css'), **css_options
      end,
      *entry.fetch(:js, []).then do |files|
        Array(files)
      end.map do |file|
        javascript_include_tag file.delete_suffix('.js'), **js_options
      end,
    ].join("\n")
  end
~~~
We used it like this:
~~~ERB
<%== asset_pack_tags(:main, js_options: { defer: true }) %>
~~~
* For cleaning up, we have this in our Makefile:
~~~Makefile
cleanup:
	rm -f ./app/assets/builds/*.digested.*
~~~

We never have these double hashing issues you mentioned.

This is probably not the migration path you are looking for. However, at least we know that every person has different contexts. I don't even know if any of these apply to Vite, but it works wonders in my projects.

aaronjensen:
Our situations are more similar than you think. The main difference is that rollup (which vite uses) will sometimes import entrypoints from chunks, which means I can't use the technique of allowing propshaft to fingerprint the entrypoints, unfortunately. It's rather annoying, but now that I have the ability to interpret the Vite manifest, it works for now. This particular app is a legacy app anyway. All our new stuff is much simpler -- import maps and regular old JavaScript without a compiler. In some cases we bundle, but that's about it.

PR #145, opened 2023-07-19
Improve log message

tmaier:
`... '#{resolved_path}' from #{resolved_path}` makes no sense, as it references the same variable twice.

We can just drop the last one from the warning


brenogazzola:
Good catch. Could you update your PR to say which file the resolved path was removed from (it should be the `logical_path`)? This would be more useful for people who are trying to track down the problem.

brenogazzola:
Due to lack of response, I'm closing this for now. If anyone else wants to handle this, the PR will be welcome.

PR #155, opened 2023-09-28
Add support for digest_class config

aduth:
This pull request proposes to add support for a `digest_class` configuration, to override the behavior of asset digesting.

**Why?**

- More control over how the digest is computed.
   - In our use-case, my interest is more in being able to shorten the length of the digest from 32 characters to 7 or 8 characters, and this is a roundabout way of doing that while offering greater overall control.
- Feature parity with [the equivalent Sprockets configuration](https://github.com/rails/sprockets/blob/5b040f3bec503ded8a98c0c911c7a77dd1f5bf1b/lib/sprockets.rb#L26), to streamline the process of migrating from Sprockets to Propshaft.

**Open Questions:**

- Should this have any impact on the behavior of `already_digested?`, whose implementation may vary depending on the digest logic?
- Is there a better way to "pass down" configuration options than carrying along the constructor argument as proposed here?
   - I was largely following the prior art of how `version` was passed along
- If there's not an interest in adding this configuration, would there be more reception to either (a) allowing the digest length to be configuration, or (b) changing the default to a shorter length? Either of these would satisfy at least my own intended use-case.

### DIFF
--- a/lib/propshaft/asset.rb
+++ b/lib/propshaft/asset.rb
@@ -2,10 +2,13 @@ require "digest/sha1"
 require "action_dispatch/http/mime_type"
 
 class Propshaft::Asset
-  attr_reader :path, :logical_path, :version
+  attr_reader :path, :logical_path, :digest_class, :version
 
-  def initialize(path, logical_path:, version: nil)
-    @path, @logical_path, @version = path, Pathname.new(logical_path), version
+  def initialize(path, logical_path:, digest_class: Digest::SHA1, version: nil)
+    @path         = path
+    @logical_path = Pathname.new(logical_path)
+    @digest_class = digest_class
+    @version      = version
   end
 
   def content
@@ -21,7 +24,7 @@ class Propshaft::Asset
   end
 
   def digest
-    @digest ||= Digest::SHA1.hexdigest("#{content}#{version}")
+    @digest ||= digest_class.hexdigest("#{content}#{version}")
   end
 
   def digested_path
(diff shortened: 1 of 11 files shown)


lib/propshaft/asset.rb
```diff
@@ -21,7 +24,7 @@ def length
   end
 
   def digest
-    @digest ||= Digest::SHA1.hexdigest("#{content}#{version}")
+    @digest ||= digest_class.hexdigest("#{content}#{version}")
```
zachmargolis, line 27: Do you think that this would be more flexible/clear if `digest_class` was a `Proc`? I know that the sprockets config was `digest_class`, but if we wanted to provide custom behavior, we'd have to dig in to the source to find out that we call `#hexdigest` on it, but a proc might have a more obvious interface?

```suggestion
    @digest ||= digester.call("#{content}#{version}")
```

and something like in the default configs
```ruby
digester: ->(content_version) { Digest::SHA1.hexdigest(content_version) }
```

aduth: I think if this were a greenfield feature, I'd agree and prefer something like what you've proposed.

In this case, I definitely leaned more on the Sprockets compatibility angle, even if it ends up being more clunky to use in our specific use-case.

I'd be curious for the maintainer's take on whether Sprockets compatibility is a desirable approach, vs. rethinking how best this option might be exposed.

brenogazzola:
@aduth Thank you for the PR, however one of the reasons we've built this gem was to have a clean slate to work on compared to Sprockets. This means that feature parity with Sprockets is the opposite of what we want. We can revisit this at a later date, whoever as long was Sprockets is supported, features that are not core to Propshaft (full support for no build, better support for builders) are not our focus.

Issue #158, opened 2023-10-13
Revisit Gzip compression support?


nickjj:
Hi,

Sorry for bringing this up again because someone opened https://github.com/rails/propshaft/issues/86 about ~18 months ago but I think the landscape of things may have changed since then, especially after seeing DHH's keynote at RailsWorld 2023.

I'd like to quote my comment in that issue:

> NGINX has a cost to do compression since it takes CPU time to do compression on the fly, especially if it's maxed out. This is objectively worse than having your digester compress the files for you beforehand, then you can configure NGINX to say the files are already compressed and serve them as they are without compute time.
>
> Tools like Tailwind and esbuild do not compress files either which leaves performing this action up to the user. That means every user will need to invent their own way to do this and also apply it for every deployment. Would it be fair to say this goes against the grain of what Rails tries to provide? Especially so with Rails pushing hard to use import maps and drop front-end toolchains.

DHH mentioned he likes that Propshaft is much smaller than Sprockets and only really focuses on 2 things but I think given the above having compression support built in aligns well with compressing complexity?

This also aligns with Kamal's philosophy of getting developers deploying their applications. Not everyone is going to be using a CDN where compression sometimes happens. Deploying to a single server with NGINX in front is a really popular use case, personally I've been doing that for around 10 years now (~8 of them with Docker).

I think adding support for Gzip should at least be considered. The `zlib` module is [built into Ruby](https://ruby-doc.org/3.2.0/exts/zlib/Zlib/GzipWriter.html). I know the feature is more than just the few lines of code to create a `compress_file` function, but just throwing it out there.

Additionally, adding Brotli is another option, in the keynote DHH mentioned to skate to where the puck is going and Brotli is probably where it'll be. It's well supported by browsers but NGINX doesn't support it out of the box. It's an NGINX Plus feature although you can compile NGINX yourself with the [Brotli module](https://github.com/google/ngx_brotli). Still, you could create the files with Propshaft and let users take it from there. You could technically support both compression methods.

brenogazzola:
Thank you for the well reasoned arguments @nickjj. We can revisit this, but here's a bit more of our reasoning to so far not support compression:

[...]

PR #161, opened 2023-10-31
Document how to enhance `assets:precompile` to provide compressed assets

rmacklin:
https://github.com/rails/propshaft/issues/158 provided some good discussion and a code example for how to provide compressed assets. This seems like useful information to add to the documentation, especially for folks who want to use Propshaft and are running their application behind NGINX with the `gzip_static` configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,29 @@ On the other hand, if you're already bundling JavaScript and CSS through a Node-
 But for greenfield apps using the default import-map approach, Propshaft can also work well, if you're able to deal with vanilla CSS.
 
 
+## Providing compressed versions of assets
+
+If you're not using a CDN that will automatically serve compressed versions of the compiled assets, you may want your web server to do so. You could let your web server compress assets on the fly, or to minimize overhead you could configure it to serve already-compressed files from disk (e.g. through NGINX's `gzip_static` directive).
+In order to provide those compressed asset files (if your existing build configuration doesn't already create them), you can enhance the `assets:precompile` rake task to create the compressed files. For example, to create Brotli-compressed versions of each file, you could add this to your application (e.g. inside an initializer):
+
+```rb
+Rake::Task["assets:precompile"].enhance do
+  assembly    = Rails.application.assets
+  output_path = assembly.config.output_path
+
+  assembly.load_path.assets.each do |asset|
+    asset_path      = output_path.join(asset.digested_path)
+    compressed_path = output_path.join(asset.digested_path.to_s + ".br")
+
+    unless compressed_path.exist?
+      Propshaft.logger.info "Compressing #{asset.digested_path}"
+      `brotli #{asset_path} -o #{compressed_path}`
+    end
+  end
+end if Rake::Task.task_defined?("assets:precompile")
+```
+
+
 ## Will Propshaft replace Sprockets as the Rails default?
 
 Most likely, but Sprockets need to be supported as well for a long time to come. Plenty of apps and gems were built on Sprocket features, and they won't be migrating soon. Still working out the compatibility story. This is very much beta software at the moment.


nickjj:
Nice one.

What do you think about changing the example to use gzip instead of Brotli since it's more supported out of the box by NGINX and other web servers? It could probably be a zero dependency "production ready" solution that uses: https://ruby-doc.org/3.2.0/exts/zlib/Zlib/GzipWriter.html

rmacklin:
No strong preference from me. Would you mind using the "Add a suggestion" feature to suggest the specific changes you had in mind using `GzipWriter` (especially if you're able to verify the code locally - I'm on mobile at the moment)?

README.md
```diff
@@ -47,6 +47,29 @@ On the other hand, if you're already bundling JavaScript and CSS through a Node-
...
+  assembly    = Rails.application.assets
+  output_path = assembly.config.output_path
+
+  assembly.load_path.assets.each do |asset|
+    asset_path      = output_path.join(asset.digested_path)
+    compressed_path = output_path.join(asset.digested_path.to_s + ".br")
+
+    unless compressed_path.exist?
+      Propshaft.logger.info "Compressing #{asset.digested_path}"
+      `brotli #{asset_path} -o #{compressed_path}`
```
nickjj, line 66: What do you think about using https://ruby-doc.org/3.2.0/exts/zlib/Zlib/GzipWriter.html instead of Brotli since more web servers support GZip out of the box vs Brotli? Then the docs could act as a copy / paste'able production ready example that would work for most folks.

Perhaps the `brotli ...` line could even remain commented out to show how you could do it for both?

nickjj:
> Would you mind using the "Add a suggestion" feature to suggest the specific changes you had in mind using GzipWriter (especially if you're able to verify the code locally - I'm on mobile at the moment)?

No problem. I can provide confirmation on if it works or not tomorrow morning EST.

brenogazzola:
Could you move this to the upgrade guide please? 

Compressions is a feature we believe should be handled by something other than propshaft, so this could was just a suggestion for a specific use case and therefore not something we want to put emphasis (which placing on the README does)

rmacklin:
I actually considered putting this in the "Upgrading from Sprockets to Propshaft" guide, but I hesitated because even though Sprockets provided this functionality, it's still useful to people who are starting a new application with Propshaft, and those people wouldn't have any reason to look at the "Upgrading from Sprockets to Propshaft" guide.

Is there another place besides that guide and the README where this could go? Do we want to add a new document, like `FAQ.md` or something else? Alternatively, we could keep this in the README but replace the inline code sample with a link to this comment: https://github.com/rails/propshaft/issues/158#issuecomment-1786039189 - that would prevent this section from taking too large a percentage of the README.

nickjj:
When running this code in production mode with Sprockets, we get this error after the assets get digested:

```
6.710 NoMethodError: undefined method `config' for nil:NilClass
6.710 /app/config/initializers/precompile.rb:3:in `block in <main>'
```

Line 3 is: `output_path = assembly.config.output_path`, assembly is defined on line 2 with `assembly = Rails.application.assets`.

Steps to reproduce in a repeatable project:

- Clone and follow the readme in https://github.com/nickjj/docker-rails-example
- Add this to `config/initializers/precompile.rb`:

```rb
Rake::Task["assets:precompile"].enhance do
  assembly    = Rails.application.assets
  output_path = assembly.config.output_path

  assembly.load_path.assets.each do |asset|
    asset_path      = output_path.join(asset.digested_path)
    compressed_path = output_path.join(asset.digested_path.to_s + ".gz")

    unless compressed_path.exist?
      Propshaft.logger.info "Compressing #{asset.digested_path}"

      Zlib::GzipWriter.open(compressed_path) do |gz|
        gz.mtime     = File.mtime(asset_path)
        gz.orig_name = asset_path
        gz.write IO.binread(asset_path)
      end
    end
  end
end if Rake::Task.task_defined?("assets:precompile")
```

- In the `.env` file, set these env vars (they are already defined but with development settings) and `docker compose build`:

```
export COMPOSE_PROFILES=postgres,redis,web,worker,cable
export RAILS_ENV=production
export NODE_ENV=production
export RAILS_SERVE_STATIC_FILES=true
```

The error will happen during the build process.

brenogazzola:
> When running this code in production mode with sprockets (propshaft isn't installed)

@nickjj This code is specific to Propshaft, which does not have a compression feature and should not be used in a project using Sprockets, which has that feature.

brenogazzola:
@rmacklin How about we take take a page out of Sprockets book and create a "guides" folder? Then we can move the upgrade guide to there too and create an "extending_propshaft.md" file and have this code and explanation there. Could also move the notice about the file watcher that is the readme me to there. And finally edit the readme to point to those guides.

nickjj:
> This code is specific to Propshaft, which does not have a compression feature and should not be used in a project using Sprockets, which has that feature.

Right but code is failing on `assembly    = Rails.application.assets` because it's coming back as nil. Is `Rails.application.assets` something Propshaft adds to Rails?

brenogazzola:
If you are asking this just as a curiosity, then `.assets` variable is something that both Sprockets and Propshaft assign a value to during application boot. My best guess as to why you are getting a null pointer with Sprockets but it works with Propshaft, is that Sprockets sets it later than Propshaft.

You could probably use Rails `to_prepare` or `after_initialize` to fix the null pointer, but then you'd just get another error, since the object Sprockets assigns to `assets` does not have a `load_path` method.

nickjj:
I see. I don't want to clutter this PR with unrelated comments but Propshaft's upgrade guide mentions deleting the `assets.rb` initializer but my initializer had `Rails.application.config.assets.paths << "/node_modules"` and the Propshaft docs don't mention how to add new load paths.

I'll need to make that customization with Propshaft to be able to test this PR.

brenogazzola:
If you are using node_modules, you are probably using a bundler like webpack/esbuild. 

In this cases there's not need to add node_modules to the path. Instead, simply have your bundler place the compiled assets (css/js) into `app/assets/builds`, and propshaft will pick them up automatically. 

It's pretty much what the bundling gems (`jsbundling-rails` and `cssbundling-rails`) do.

nickjj:
I'm using esbuild and it's configured to place assets into `/node_modules`. It's important that the path doesn't change because it's a non-volume mounted location of the file system within Docker.

How can I configure Propshaft to include `/node_modules` as an assets path similar to sprockets?

brenogazzola:
Don’t do that. Propshaft paths do not work like Sprockets paths. 

Anything you add to paths will be copied to the `public/assets` folder by Propshaft. I doubt you want the entirety of the `node_modules` folder there, as that would be thousands of files and hundreds of megabytes, and it would massively slow down the precompile.

Like I said, have your bundler, which uses `node_modules` folder as the source of the libs for your own js files, place the files it generates in `app/assets/builds` and from there propshaft will copy them to `public/assets`. That’s the correct workflow to integrate Propshaft with bundlers like webpack/esbuild/etc.

brenogazzola:
That said, if you really want to do that to see what happens, instead of removing the `assets.rb` file, place this line in it:

```ruby
Rails.application.config.assets.paths.unshift Rails.root.join("node_modules").to_s
```

nickjj:
Just to make sure I'm 100% following here. Here's what I do today with sprockets and everything is working. Everything gets bundled, digested and gzipped as expected and my bundles only include my app's custom JS and CSS:

- I configured esbuild to put my bundles in `app/assets/builds` for JS and CSS
- I configured `.yarnrc` with `--modules-folder /node_modules` so that any JS dependencies I install end up in `/node_modules`
- I configured Rails with `Rails.application.config.assets.paths << "/node_modules"` so that whatever tool needs this path (Rails, sprockets, etc.?) knows assets exist there and not in the relative `node_modules/` directory that's the default

With Propshaft it sounds like step 1 doesn't change. Step 2 can't change because that's where I Yarn install everything to. The part I'm not clear on is step 3. The upgrade guide says to delete the initializer where this step was previously included but it has no alternative. Maybe this path configuration isn't needed at all with Propshaft and this whole concept goes away? In other words I'm good to go out of the box with my old steps 1 and 2 with Propshaft?

brenogazzola:
Correct, you don’t need step 3. Propshaft does not care about node modules. It’s esbuild responsibility to read stuff from there when compiling your files. Propshaft only cares about where your compiled files will be places and that’s the builds folder.

brenogazzola:
@nickjj Need to change the if in that code I gave you:
```
Rake::Task["assets:precompile"].enhance do
  # Code here
end if defined?(Rack) && Rake::Task.task_defined?("assets:precompile")

```

dhh:
We are working on a new solution to provide compression out of the box without CDN. See the Airlock issue on the Rails 8 milestone. I think that's a better path.

nickjj:
Would it still be possible to provide a working initializer here in case Airlock doesn't solve this use case?

Based on [the milestone](https://github.com/rails/rails/issues/50479) it sounds like it would compress assets on the fly similar to how nginx would compress them, but the original issue for this use case was about pre-compressing them because it's more efficient. The original issue is here btw: https://github.com/rails/propshaft/issues/158

dhh:
I don’t think that’ll be a concern of high enough importance for most to bother with. Airlock will only compress once per asset, though it will do so on demand. But computers are fast enough that it isn’t a concern I think most people should have to even think about.

I want to keep Propshaft as lean as possible. That includes which seeds to plant in people’s minds about what’s needed or not when reading the docs.

nickjj:
Ok thanks, I'll drop a request in the Rails forums to see if we can figure out debugging this almost-working-initializer for folks who plan to stick with using nginx which doesn't only compress assets once.

morgoth:
I understand to not include it in the readme, but maybe it would be helpful to put it in the upgrade guide in section about migrating from sprockets.
@dhh what do you think?

dhh:
I think it'll all be moot because Thruster will be the default in Rails 8, and that'll provide the compression needed.

nickjj:
Now that Rails 8.0 is out and Propshaft is the default, can we come up with a working snippet that's documented in https://github.com/rails/propshaft/pull/161#issuecomment-1788304229? That comment and a few comments below it explain the errors it produces and I've hit a dead-end on troubleshooting it.

There are folks out there not using Thruster and would like to continue having nginx serve pre-compiled assets but still take advantage of Propshaft.

PR #162, opened 2023-11-06
Upgrade VS Code extension to Ruby LSP

esasse:
This upgrades VS Code extension in DevContainer to the new Ruby LSP. It also updates settings file format.


brenogazzola:
@rafaelfranca Mind taking a look at this one? I don't use VSCode, so not sure of the impact.

esasse:
@brenogazzola if it helps, take a look at main rails repo, they are using this same extension: https://github.com/rails/rails/blob/main/.devcontainer/devcontainer.json#L36.

PR #163, opened 2023-11-07
Improve source mapping compiler log message

esasse:
Improves previous PR https://github.com/rails/propshaft/pull/145.

Now the `logical_path` is properly logged, making it easier to find the problem.

### DIFF
--- a/lib/propshaft/compiler/source_mapping_urls.rb
+++ b/lib/propshaft/compiler/source_mapping_urls.rb
@@ -6,7 +6,7 @@ class Propshaft::Compiler::SourceMappingUrls < Propshaft::Compiler
   SOURCE_MAPPING_PATTERN = %r{(//|/\*)# sourceMappingURL=(.+\.map)(\s*?\*\/)?\s*?\Z}
 
   def compile(logical_path, input)
-    input.gsub(SOURCE_MAPPING_PATTERN) { source_mapping_url(asset_path($2, logical_path), $1, $3) }
+    input.gsub(SOURCE_MAPPING_PATTERN) { source_mapping_url(logical_path, asset_path($2, logical_path), $1, $3) }
   end
 
   private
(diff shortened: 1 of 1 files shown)


lib/propshaft/compiler/source_mapping_urls.rb
```diff
@@ -6,11 +6,14 @@ class Propshaft::Compiler::SourceMappingUrls < Propshaft::Compiler
   SOURCE_MAPPING_PATTERN = %r{(//|/\*)# sourceMappingURL=(.+\.map)(\s*?\*\/)?\s*?\Z}
 
   def compile(logical_path, input)
-    input.gsub(SOURCE_MAPPING_PATTERN) { source_mapping_url(asset_path($2, logical_path), $1, $3) }
+    @logical_path = logical_path
+    input.gsub(SOURCE_MAPPING_PATTERN) { source_mapping_url(asset_path($2), $1, $3) }
```
brenogazzola, line 10: I'm not much of a fan of setting instance variables in the middle of methods to avoid passing them forward. Specially because the other two methods that use it still need other variables passed as parameters. Better to repeat `logical_path` here

brenogazzola (changes requested):

lib/propshaft/compiler/source_mapping_urls.rb
esasse: Thanks, I made the changes you asked. But I'd like to clarify why I decided on the previous path: `logical_path` is used in all methods of the class, so it's usually recommended to store it in an instance variable so that you don't need to pass it as parameters everywhere. And more parameters usually also make code harder to read.

esasse:
@brenogazzola any other changes you want me to do so that this PR can be merged?

brenogazzola:
Thanks, sorry for the delay!

PR #165, opened 2023-11-07
Adds the assets:clean[count] parameter

codergeek121:
Sprockets supported the parameterized `clean` Rake task, to keep the `:count` latest assets versions. IIRC webpacker supported it as well.

If you deployed with Capistrano, this was configurable via the `:keep_assets` [config](https://github.com/capistrano/rails/blob/master/lib/capistrano/tasks/assets.rake#L33).

This PR adds this to propshaft's clean task as well instead of the hardcoded `2`.


codergeek121:
Thank you :)

PR #166, opened 2023-11-13
Ensure file content is read as UTF-8

brenogazzola:
As reported in https://github.com/rails/sprockets/issues/498. Propshaft has the same bug. Will keep this unmerged for a while for testing and further opinion from others.


brenogazzola:
Propshaft writes the file correctly, even if the specific test in this PR fails without the enforcing UTF-8. Closing as not needed.

PR #167, opened 2023-11-22
Fix typo in "Asset content" section of UPGRADING.md

HashNotAdam:
When referring to how to escape HTML tags, the documentation forgot to call the `content` method.

When using `html_safe`, an exception will be raised because `html_safe` is not defined on `Propshaft::Asset`.

Using `raw`, on the other hand, will call `to_s` on the `Propshaft::Asset` instance, returning an expected tag


brenogazzola:
Thanks!

PR #168, opened 2023-11-22
Support base64 encoded digests to account for rollup 4.x

aaronjensen:
Rollup 4 changed its asset digesting to include `_` and `-` (they are now base64 url safe hashes). Without this, assets such as `chunk-_OUZQqad.digested.js` will not be served by the propshaft server.

I did not change the regex that extracts the digest that Propshaft is expected to generate as that didn't seem appropriate.

This change is necessary for compatibility with Vite 5 and rollup 4 (unless of course there's a better way to do this).

I could not find a way to revert Rollup's hashing technique.


brenogazzola:
Thanks.

aaronjensen:
@brenogazzola Is there anything that I can do to help get this ready for a release? We have been unable to run the new version without a monkey patch for a few months. Thank you.

theodorton:
@aaronjensen I'm not sure when the next release will be here, but until then you can pull the gem from Github instead of rubygems:

```rb
gem 'propshaft', github: 'rails/propshaft'
```

aaronjensen:
Thanks, @theodorton, I'm aware of that, but we try to avoid git dependencies for various reasons. Is there anything I can do to help make it easier to follow the release process for propshaft?

PR #172, opened 2023-12-24
Don't say this is beta software [skip ci]

northeastprince:
Since many production apps like [HEY have been using this](https://twitter.com/dhh/status/1491843448881692679) for years now, I think it's safe to remove this warning. Personally, I was kinda scared of using this in my apps until I saw the full story.


brenogazzola:
Thank you for the PR, but this is still beta software. While some people are using it in production, it's not version 1.0 yet and there are some problems we'd to solve before saying it's ready for everyone to use.

Issue #169, opened 2023-12-07
allow a configurable digest length


wlipa:
The asset digest is 40 random hex characters.  This is far more than needed to provide disambiguated assets and contributes to larger page size (which doesn't compress well due to the randomization) and more noisy looking HTML source.

It would be nice to have a config option to specify the digest size.  It can be monkeypatched like this, but obviously a patch is more fragile.

More boldly, I would suggest that a shorter digest size (here 8) would be the right default for almost every app.

```
class Propshaft::Asset
  def digest
    @digest ||= Digest::SHA1.hexdigest("#{content}#{version}")[0,8]
  end
end
```

wlipa:
I don't have numbers, but consider that it's every reference to every asset, including not only what's on the page but asset requests, keys in caches, etc.   Some pages might have hundreds or even thousands of assets.  And these fragments don't compress that well (basically just reversing the hex encoding).

To be honest that part of it bugs me less than the impact on the source and developer tools.  I dislike having these large hex fragments all over the place - it makes it harder to quickly look at and understand the source.

brenogazzola:
Alright, I’m not against reducing it. It did feel excessive to me at times. But I’d like to be sure we are choosing a good default. Unlike JS bundlers, propshaft is handling a lot more assets then they are, so their default might not be applicable. 

Why is 8 good enough?

wlipa:
The digest needs to be long enough to have a nearly insignificant chance that two versions of an asset will accidentally get the same digest.  8 hex is a long - 4 billion.  I consider an N in 4 billion chance that a new version of an asset will overlap one of the N previous versions in the cache to be nearly insignificant.

brenogazzola:
I've talked to the others and we've agreed to reduce the digest size. Would you like to open a PR for that?

wlipa:
Sure, I will do that.  Thanks for considering the issue.

dhh:
Let's def just pick a value. This isn't something we need to allow configuration on. 8 sounds good to me.

wlipa:
Addressed by #173

midnight-wonderer:
I hated to be that guy, but the __N in 4 billion chance__ estimation is far off.

This is the [birthday paradox](https://en.wikipedia.org/wiki/Birthday_problem).  
Here is my understanding:

The eight digits of a hexadecimal number are 32 bits long; this will have a 0.5 probability of collision at about 2^16 builds.
2^16 builds = 65,536 builds

[...]

PR #173, opened 2023-12-28
Shorter digest length

wlipa:
Addresses https://github.com/rails/propshaft/issues/169 by reducing the digest length to 8  in order to shorten asset references and make the source less noisy.


lib/propshaft/asset.rb
```diff
@@ -21,7 +21,7 @@ def length
   end
 
   def digest
-    @digest ||= Digest::SHA1.hexdigest("#{content}#{version}")
+    @digest ||= Digest::SHA1.hexdigest("#{content}#{version}")[0,8]
```
dhh, line 24: You should be able to use .first(8) here.

PR #174, opened 2024-01-03
Add Ruby 3.3 to CI matrix

m-nakamura145:
This PR adds Ruby 3.3 to CI matrix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ jobs:
           - "3.0"
           - "3.1"
           - "3.2"
+          - "3.3"
         continue-on-error: [false]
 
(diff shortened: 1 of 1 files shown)


dhh:
Looks like we need to update the bundle for this to pass.

m-nakamura145:
Thank you! It seems that [nokogiri version 1.16.0 or higher](https://rubygems.org/gems/nokogiri/versions/1.16.0-x86_64-linux) is required.

> nokogiri-1.14.0-x86_64-linux requires ruby version < 3.3.dev, >= 2.7, which is
> incompatible with the current version, 3.3.0

theodorton:
@dhh Good to merge?

Issue #176, opened 2024-01-09
Using Propshaft::Asset#content with UTF-8 encoding


denzelem:
Recently I tried to use inline styles with propshaft to render PDFs with https://github.com/Studiosity/grover. I figured out, that using `Propshaft::Asset#content` does not work well, when reading `UTF-8` encoded files. 

```
!!!
%html
  %head
    %meta(content='text/html; charset=UTF-8' http-equiv='Content-Type')
    %meta(name='viewport' content='width=device-width, initial-scale=1')
    %title= @title
    = inline_stylesheet_link_tag('letter.css')

  %body
    = yield
```
 
## Does not work
 
```
 module AssetHelper
  def inline_stylesheet_link_tag(asset_name)
    content_tag(:style, Rails.application.assets.load_path.find(asset_name).content.html_safe )
  end
end
```

```
Encoding::CompatibilityError at /letter.html
incompatible character encodings: ASCII-8BIT and UTF-8
```

## Works
  
```
 module AssetHelper
  def inline_stylesheet_link_tag(asset_name)
    content_tag(:style, File.read(Rails.application.assets.load_path.find(asset_name).path, encoding: 'UTF-8').html_safe )
  end
end
```

Do you think it make sense to e.g. add an encoding option to Propshaft::Asset#content? I can create a PR for this. Otherwise just close this issue.

brenogazzola:
Could you test the branch in this closed PR of mine? I thought this wouldn't be needed: https://github.com/rails/propshaft/pull/166.

denzelem:
> Could you test the branch in this closed PR of mine? I thought this wouldn't be needed: #166.

For my use case your PR is working. But from my technical understanding (see [this blog post](https://tenderlovemaking.com/2020/01/13/guide-to-string-encoding-in-ruby.html)) it is wrong to assume all assets are encoded with `UTF-8`.

`File.binread(path).force_encoding("UTF-8")` seems to me the same as `File.read(path, encoding: 'UTF-8')`. Both assume an asset encoded with `UTF-8`, but might return a wrong result / raise an exception for other encodings.

What do you think of adjusting your PR to something like:

```
def content(encoding: 'ASCII-8BIT')
  File.read(path, encoding: encoding)
end
```

Only the person that uses `Propshaft::Asset#content` can know which encoding the asset has. And falling back to `ASCII-8BIT'` is the best what one could do imho.

Wdyt?

denzelem:
[...]

PR #180, opened 2024-01-23
Add encoding options to Resolver::Static#read and Resolver::Dynamic#read

denzelem:
PR for https://github.com/rails/propshaft/issues/176.

### DIFF
--- a/lib/propshaft/resolver/dynamic.rb
+++ b/lib/propshaft/resolver/dynamic.rb
@@ -12,9 +12,9 @@ module Propshaft::Resolver
       end
     end
 
-    def read(logical_path)
+    def read(logical_path, options = {})
       if asset = load_path.find(logical_path)
-        asset.content
+        asset.content(**options)
       end
     end
   end
--- a/lib/propshaft/asset.rb
+++ b/lib/propshaft/asset.rb
@@ -8,8 +8,8 @@ class Propshaft::Asset
     @path, @logical_path, @version = path, Pathname.new(logical_path), version
   end
 
-  def content
-    File.binread(path)
+  def content(encoding: "ASCII-8BIT")
+    File.read(path, encoding: encoding)
   end
 
   def content_type
--- a/lib/propshaft/resolver/static.rb
+++ b/lib/propshaft/resolver/static.rb
@@ -12,9 +12,9 @@ module Propshaft::Resolver
       end
     end
 
-    def read(logical_path)
+    def read(logical_path, encoding: "ASCII-8BIT")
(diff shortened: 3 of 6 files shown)


lib/propshaft/resolver/dynamic.rb
```diff
@@ -12,9 +12,9 @@ def resolve(logical_path)
       end
     end
 
-    def read(logical_path)
+    def read(logical_path, options = {})
       if asset = load_path.find(logical_path)
-        asset.content
+        asset.content(**options)
```
brenogazzola, line 17: Is there a reason why you used `options` here instead of just `encoding` like in `assets.rb` and `static.rb`?

denzelem: I used an option hash to avoid duplicating the default value to this top-level method.

Alternative 1:

```
def read(logical_path, encoding: 'ASCII-8BIT')
  # ...
  asset.content(encoding: encoding)
  # ...
end
```

Alternative 2:

```
def read(logical_path, encoding: nil)
  # ...
  options = {}
  options[:encoding] = encoding if encoding.present?
  asset.content(**options)
  # ...
end
```

@brenogazzola Wdyt?

dhh:
Don't see how this works when the asset is a binary file? If we're assuming ASCII-8BIT, what happens when you try to read a PNG?

denzelem:
> Don't see how this works when the asset is a binary file? If we're assuming ASCII-8BIT, what happens when you try to read a PNG?

@dhh Do you think that `File.binread(path)` is different to `File.read(path, encoding: encoding)`? My assumption was, that [binread](https://github.com/ruby/ruby/blob/ruby_3_3/io.c#L12208) (rb_ascii8bit_encoding C-Method) reads the bytes with the encoding `ASCII-8BIT`.

So reading an PNG image should be no breaking change with this PR.

```
require 'open-uri'
require 'tmpdir'

image_url = 'https://picsum.photos/id/237/200/300'

Dir.mktmpdir do |dir|
  image_path = File.join(dir, 'example.jpg')

  URI.open(image_url) do |image|
    File.open(image_path, 'wb') do |file|
      file.write(image.read)
    end
  end

  puts 'Encoding binread ' + File.binread(image_path).encoding.to_s
  puts 'Encoding read with ASCII-8BIT ' + File.read(image_path, encoding: 'ASCII-8BIT').encoding.to_s
end

exit
```

```
Encoding binread ASCII-8BIT
Encoding read with ASCII-8BIT ASCII-8BIT
```

theodorton:
@dhh @brenogazzola I don't see any blockers for merging this - what do you think?

Issue #123, opened 2022-12-20
Asset digest is computed before compilation


grncdr:
This is the underlying cause of #90 but the description and comments there conflate a number of issues.

## Steps to reproduce

Create two CSS files:

```css
/* application.css */
import('other.css');
```

```css
/* other.css */
body {
  background-color: pink;
}
```

Load `application.css`:

```erb
<%= stylesheet_link_tag 'application' %>
```

Load the page (with browser caching enabled) and observe a pink background.

Change the body background in `other.css`:

```diff
-  background-color: pink;
+  background-color: red;
```

Reload the page.

## Expected result

The background is now red.

## Actual result

The background is still pink.

## Why this happens

The digest for `application.css` is computed from the unprocessed file content. This causes `Propshaft::Asset` to produce the _same_ digest for the asset even though the actual contents at that URL will be different (due to the changing digest of `other.css`).

Because the asset is served with a very long expiry, browsers will hang on to the outdated version of the asset for a very long time.

## Why that's a problem

In a production scenario, browsers would never get the new version of `other.css`. This makes using `@import` a pretty big foot gun.

grncdr:
Just wanted to come back to this issue and confirm that it definitely affects production deployments, because the browser will use a cached version of the stylesheet with older content hashes for the imported files.

We are currently working around the issue by adding a cache-busting comment to the "root" stylesheet before running assets:precompile in CI, but a more robust solution from propshaft would be very appreciated.

dhh:
@brenogazzola We'd need to introduce two digests to deal with this, yeah? Basically a raw source digest and a processed source digest. And then keep those straight for when to use.

brenogazzola:
@dhh I think it's a bit more complicated than that? What about multiple levels of import? 

```css
/* application.css */
import('elements.css');
```
```css
/* elements.css */
import('avatar');
import('button');
import('form');
```
```css
/* button.css */
.button {
  background-color: red;
}
```

If you change `red` in `button.css` than the raw of button changes, but not of elements, which means that the processed of elements changes, but the processed of application.css does not.

I can think of two options:
1 - Keep digesting/redigesting until no digests change
2 - Teach propshaft digest/compile things in order

grncdr:
> but the processed of application would not (since none of its imports had changes in the raw digest)?

[...]

PR #188, opened 2024-05-17
Digests of files that can have dependencies on other files in the load path need to reflect those dependencies

dhh:
CSS files that reference other CSS files via `@import` or images via `url()` need to have their digest reflect changes in the dependencies. This PR adds that by including the content of all dependencies in the calculation for such assets with dependencies.

This takes some inspiration from https://github.com/rails/propshaft/pull/175 but avoids the dependency tree tracking.

### DIFF
--- a/benchmarks/css_asset_urls
+++ b/benchmarks/css_asset_urls
@@ -25,5 +25,5 @@ compiler = Propshaft::Compiler::CssAssetUrls.new(assembly)
 
 Benchmark.ips do |x|
   x.config(time: 5, warmup: 2)
-  x.report("compile") { compiler.compile(asset.logical_path, asset.content) }
+  x.report("compile") { compiler.compile(asset, asset.content) }
 end
--- a/lib/propshaft/assembly.rb
+++ b/lib/propshaft/assembly.rb
@@ -15,7 +15,7 @@ class Propshaft::Assembly
   end
 
   def load_path
-    @load_path ||= Propshaft::LoadPath.new(config.paths, version: config.version)
+    @load_path ||= Propshaft::LoadPath.new(config.paths, compilers: compilers, version: config.version)
   end
 
   def resolver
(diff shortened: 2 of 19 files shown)


wlipa:
For some reason, I'm not getting digested assets in CSS files with this.  I do get digests from the asset helpers though.

dhh:
@wlipa What do you mean exactly? That `@import url` isn't being replaced with a digested url?

dhh:
Ah, I see the problem.

wlipa:
In content_with_compile_dependencies, the references will need to be returned in a stable order in order for digests to be stable.  Maybe they are if it's based on source code order?  Not 100% sure on merging.

dhh:
Should be stable given the scan will always return them in order. But good question about set guaranteeing a stable order. Will investigate.

dhh:
As far as I can tell, since Ruby 2.5, the insert order is guaranteed in a set, and that includes merging. So we should be good.

wlipa:
This is checking out for me now.  👍🏼

dhh:
Thanks for your help with this, @wlipa. Got some good inspiration for this approach from your work, and I think we ended up in a nice place without complicating the domain model much.

dhh:
Fixes https://github.com/rails/propshaft/issues/123 and https://github.com/rails/propshaft/issues/90. Closes #175.

wlipa:
Yeah sounds good.  It's definitely simpler.  The only thing I noticed that's a bit of a trade-off is reading all the dependent assets into memory to do that sum.  Could be a lot with big images.  But maybe it's not that much of a problem in practice.

dhh:
Yeah but we read them into memory anyway to digest them individually. Although of course it's perhaps a higher peak if a single css file references a ton of images. Still, doubt it's an issue. Your images shouldn't be that big in the first place if referenced by css!

PR #190, opened 2024-05-19
Provide :all and :app helpers for stylesheet references via cached methods

dhh:
Add caching to `stylesheet_link_tag :all` so we don't search the load path every time when the files haven't changed in production.

Add `stylesheet_link_tag :app` to grab just the css files from app/assets/stylesheets/**/*.css and not all the potential engine load paths that have been added as well.

### DIFF
--- a/lib/propshaft/helper.rb
+++ b/lib/propshaft/helper.rb
@@ -6,8 +6,11 @@ module Propshaft
 
     # Add an option to call `stylesheet_link_tag` with `:all` to include every css file found on the load path.
     def stylesheet_link_tag(*sources, **options)
-      if sources.first == :all
-        super(*all_stylesheets_paths, **options)
+      case sources.first
+      when :all
+        super(*all_stylesheets_paths , **options)
+      when :app
+        super(*app_stylesheets_paths , **options)
       else
         super
       end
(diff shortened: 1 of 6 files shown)


lib/propshaft/helper.rb
```diff
@@ -6,20 +6,24 @@ def compute_asset_path(path, options = {})
 
     # Add an option to call `stylesheet_link_tag` with `:all` to include every css file found on the load path.
```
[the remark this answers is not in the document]
dhh: Yes, good point

lib/propshaft/helper.rb
```diff
@@ -6,20 +6,24 @@ def compute_asset_path(path, options = {})
...
-        .assets(content_types: [ Mime::EXTENSION_LOOKUP["css"] ])
-        .collect { |css| css.logical_path.to_s }
-        .sort
-        .uniq
+      Rails.application.assets.load_path.asset_paths_by_type("css")
+    end
+
+    # Returns a sorted and unique array of logical paths for all stylesheets in app/assets/**/*.css.
+    def app_stylesheets_paths
+      Rails.application.assets.load_path.asset_paths_by_glob("#{Rails.root.join("app/assets")}/**/*.css")
```
brunoprietog, line 26: Since this is only intended for loading styles, is there any reason to look in folders other than just stylesheets?

dhh: This was needed to get a single-line matcher for fnmatch that works with both nested directories and not. Don't think it really matters, and it does provide slightly more flexibility if someone organizes their app/assets differently.

PR #201, opened 2024-08-30
Clean old sourcemap assets

marcelolx:
This PR https://github.com/rails/propshaft/pull/171 made propshaft to maintain sourcemap files, but they should be cleaned up the same way as normal assets are.

### DIFF
--- a/lib/propshaft/output_path.rb
+++ b/lib/propshaft/output_path.rb
@@ -53,7 +53,7 @@ class Propshaft::OutputPath
     end
 
     def extract_path_and_digest(digested_path)
-      digest = digested_path.to_s[/-([0-9a-f]{7,128})\.(?!digested)[^.]+\z/, 1]
+      digest = digested_path.to_s[/-([0-9a-f]{7,128})\.(?!digested)/, 1]
       path = digest ? digested_path.sub("-#{digest}", "") : digested_path
 
       [path, digest]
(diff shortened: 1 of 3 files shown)


lib/propshaft/output_path.rb
```diff
@@ -53,7 +53,7 @@ def all_files_from_tree(path)
     end
 
     def extract_path_and_digest(digested_path)
-      digest = digested_path.to_s[/-([0-9a-f]{7,128})\.(?!digested)[^.]+\z/, 1]
+      digest = digested_path.to_s[/-([0-9a-f]{7,128})\.(?!digested)/, 1]
```
theodorton, line 56: This regex should be the same as in `server.rb`:

https://github.com/rails/propshaft/blob/af557fbd10dc9856c943b3324f340a87295f5041/lib/propshaft/server.rb#L38

Might be worth defining this as a constant on `Propshaft::Asset`.

marcelolx: Is there a reason why the server one shouldn't be the same as this one? The PR that introduced support for multiple extensions after `.digested` has made the same change to the regex made here   https://github.com/rails/propshaft/pull/14/files#diff-cea8115669d1b6f7ed0b417671e6b929c07d4f41b4e009b82182221f38b46955L45

I agree a constant might be worth it, so when you change it, all places use the new version

marcelolx: Just not sure if we should use the same regex in all places (a few different places are defining a regex like this )

PR #202, opened 2024-09-04
Add release workflow using Trusted publishing

rafaelfranca:
Releasing is done now by publishing a new release on GitHub.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,31 @@
+name: Release
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  release:
+    permissions:
+      contents: write
+      id-token: write
+
+    environment: release
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        bundler-cache: true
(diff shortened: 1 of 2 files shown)


dhh:
So flow is:

1. Run bin/release to bump version + create tag
2. Manually add a new release on GH
3. This flow pushes the gem?

dhh:
Maybe I'm not entirely following what this would buy us, then? If you still have to run bin/release, what extra do we get from having GH handle that part of the release? The gem release is still gated by RubyGems and 2FA. So security shouldn't be a consideration there?

rafaelfranca:
For small gems this doesn't buy us much other than we probably can decrease the number of people allowed to push gems in Rubygem.org.

And, given new people release access is as easy as only adding to GitHub. No need to also go through all the gems and adding that person with push access on RubyGems.org. This doesn't happen often, but when it does, sometimes leave us open to attacks.

I'm not sold the overhead of two steps is good enough for small gems, that is why I opened the PR. For Rails, I think it is no brainer.

dhh:
Yeah, I think that's probably fair. That user management is easier on GitHub than through RubyGems, but that does seem like a fairly minor advantage in the grand scheme of things. And yes, I think the real test will be whether the GH flow allows us to encode the much more complicated multi-stage release process for Rails. Would very much like to see that encoded and clarified!

PR #203, opened 2024-09-06
Ensure manifest_path is read from config properly by Assembly

keiththomps:
Ran into the following error attempting to utilize propshaft main (caused by simple typo):

image

Test likely isn't necessary to keep, but I needed it to reproduce the issue.

### DIFF
--- a/lib/propshaft/assembly.rb
+++ b/lib/propshaft/assembly.rb
@@ -32,7 +32,7 @@ class Propshaft::Assembly
 
   def processor
     Propshaft::Processor.new \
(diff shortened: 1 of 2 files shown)


tapster:
Just hit the same issue. Confirmed Keith's fix in this PR resolves 👍 . thanks

```
2.943 bin/rails aborted!
2.945 NoMethodError: undefined method `manifest_path' for an instance of Pathname (NoMethodError)
2.945
2.945 load_path: load_path, output_path: config.output_path, compilers: compilers, manifest_path: config.manifest_path.manifest_path
2.945                                                                                                                  ^^^^^^^^^^^^^^
```

PR #206, opened 2024-09-22
Prioritize app assets over engine assets

brunoprietog:
Make the app assets always have priority when resolving them, if there is another asset with the same name in a gem. For example:

```bash
bin/importmap pin trix
```

You would have a `vendor/javascript/trix.js` file. If you call "trix" from importmap, it would still take the one from Action Text gem instead. Not anymore.

### DIFF
--- a/lib/propshaft/railtie.rb
+++ b/lib/propshaft/railtie.rb
@@ -32,6 +32,7 @@ module Propshaft
     end
 
     config.after_initialize do |app|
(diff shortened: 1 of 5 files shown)


theodorton:
> Not anymore.

Are you seeing a regression compared to sprockets after migrating to propshaft?

I think we need to consider warning users in cases of ambiguous asset names, and also be clear/explicit in the documentation and configuration about load order.

Is the specific engine you're using provided from a public gem?

PR #209, opened 2024-09-30
[Regression] Use stable sort when prioritizing app assets

excid3:
This fixes a regression introduced in #206. 

`sort_by!` doesn't preserve the original order so this loses the original order. https://ruby-doc.org/3.3.5/Array.html#method-i-sort_by-21
> For duplicates returned by the block, the ordering is indeterminate, and may be unstable.

I ran into this using esbuild and had `app/assets/builds` as the first entry in paths. After #206, `app/javascript` always ended up higher priority than `app/assets/builds` so it was finding the wrong file.

### DIFF
--- a/lib/propshaft/railtie.rb
+++ b/lib/propshaft/railtie.rb
@@ -33,7 +33,7 @@ module Propshaft
 
     config.after_initialize do |app|
       # Prioritize assets from within the application over assets of the same path from engines/gems.
-      config.assets.paths.sort_by! { |path| path.to_s.start_with?(Rails.root.to_s) ? 0 : 1 }
+      config.assets.paths.sort_by!.with_index { |path, i| [path.to_s.start_with?(Rails.root.to_s) ? 0 : 1, i] }
 
       config.assets.relative_url_root ||= app.config.relative_url_root
       config.assets.output_path ||=


brenogazzola:
Thanks Chris! I was just talking about this one with a colleague who had this problem. Will just ask him to check and merge if it solves his problem too.

excid3:
Awesome, thanks for the quick review. It was certainly a head scratcher today as I was testing our Rails 8 migration!

excid3:
Alternative implementation I considered was to `partition` into two arrays and re-assign but seemed like it wasn't much clearer than `sort_by!.with_index`.

```ruby
config.assets.paths = config.assets.paths.partition { |path| path.to_s.start_with?(Rails.root.to_s) }.flatten
```

brenogazzola:
The `partition` version looks more elegant, though a `sort_by!` seems more intuitive. Any specific reason you chose one over the other?

brenogazzola:
Just for future reference, it seems that if all elements of the array are a `String`, the ordering is not lost. But if you have `Pathname` variables in there, then the order changes.

excid3:
I figured partition would allocate a couple arrays and join them back so it is probably slower. Although the arrays probably aren't large enough to matter.

excid3:
`Path` objects seem pretty common because that's what you get from `Rails.root.join`

The ones causing issues for me are from importmap-rails: https://github.com/rails/importmap-rails/blob/main/lib/importmap/engine.rb#L46-L47

Same approach in most engines/railties I think: https://github.com/rails/mission_control-jobs/blob/main/lib/mission_control/jobs/engine.rb#L93-L94

brenogazzola:
Yeah, I doubt it will make that much of a difference. But the `sort_by` is a bit more obvious I think.

PR #220, opened 2024-11-20
Add quotes to etags

zarqman:
This PR fixes etag generation in Propshaft::Server.

In contrast to other headers, etags must be quoted. (See [RFC 9110 §8.8.3](https://www.rfc-editor.org/rfc/rfc9110.html#name-etag).)

Before:  
`etag: fa74dbac`
After:   
`etag: "fa74dbac"`

### DIFF
--- a/lib/propshaft/server.rb
+++ b/lib/propshaft/server.rb
@@ -18,7 +18,7 @@ class Propshaft::Server
           Rack::CONTENT_LENGTH  => compiled_content.length.to_s,
           Rack::CONTENT_TYPE    => asset.content_type.to_s,
           VARY                  => "Accept-Encoding",
-          Rack::ETAG            => asset.digest,
+          Rack::ETAG            => "\"#{asset.digest}\"",
           Rack::CACHE_CONTROL   => "public, max-age=31536000, immutable"
         },
         [ compiled_content ]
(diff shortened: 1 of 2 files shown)


theodorton (approved):

brenogazzola:
Not sure if we should add a W/ before to make it a weak tag, but we can discuss this later. Merged.

zarqman:
My understanding is that strong etags are appropriate when: 
a) the etag is always assured to change if/when the content changes, and 
b) the strong etag is only used for a single encoding of the content (identity, gzip, brotli, etc).

In the case of (b), strong etags may still be used with different encodings as long as the etag itself also changes. With digest-based etags, that can be as simple as adding a suffix, like "originalDigest-gzip", although a fresh digest of the compressed content would be valid too (it's just hard to match a fresh digest back to the original if needed).

If an etag is reused across encodings, then it should be weak (or converted to weak). As a general rule, this conversion should be the responsibility of whatever's transforming the encodings.

Strong etags can allow for content to be cached more often or for longer, and are desirable for things like static assets.

Since propshaft doesn't involve itself with compression or any other alternate encodings, and since the digest-sourced etag will definitely change if the content changes, a strong etag seems appropriate.

brenogazzola:
Your final paragraph addresses the thing I had been wondering: CDNs that automatically serve optimized versions of image files.

I agree that since it’s not Propshaft making that change, then Propshaft can use strong tags and the CDN can handle their part.

PR #229, opened 2025-01-07
Refactor to allow digest algorithm patching.

midnight-wonderer:
Hi,
I once tried to discuss the default digest length [on #169](https://github.com/rails/propshaft/issues/169), to no avail.
How about simply refactoring things around without behavioral changes?  

I extracted
~~~ruby
Digest::SHA1.hexdigest(text).first(8)
~~~
into a private method, allowing a stable way for monkey-patching without affecting the functionality.

This keeps the current algorithm and output length, the change is minimal, and no change to any of the existing tests.

### DIFF
--- a/lib/propshaft/asset.rb
+++ b/lib/propshaft/asset.rb
@@ -30,7 +30,7 @@ class Propshaft::Asset
   end
 
   def digest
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Having private methods as a way to extend the library via monkey patch isn't something we do, and having this configurable is overkill. I think we will keep using 8 digits. Thank you for the pill request

PR #230, opened 2025-01-21
Update README.md

gobijan:
Add reference to the Rubydoc API documentation to make it easier to navigate.


brenogazzola:
Thank you for your contribution, but none of the rails gems (old or new) include a such a reference in the README file, so I'll follow their lead and say it's not needed here either.

gobijan:
Maybe this should be changed throughout the rails gems too?
From a users perspective it’s clearly better to have a direct reference to the API documentation. This is quite a normal thing in other ecosystems such as Go etc.
I just found myself repeatedly jumping back and forth and it was annoying. I also don’t want to be contributor or whatever. It’s just the right thing to do in order to reduce friction. What would be the right channel to discuss?

brenogazzola:
Two ways I’ve used in the past:

1. Open a PR on rails, post to the forum that has more visibility with a link to the PR
2. Join the discord channel and ask about it on the contribution channel.

Issue #140, opened 2023-05-29
assets:clean task is not cleaning predigested assets with `.digested` in the name


tsrivishnu:
Whenever `rake assets:clean` is run, the assets in `public/assets` with `.digested` in their name are not cleaned at all. This causes assets to accumulate on production and drastic increase in disk usage over time. 

I realised that this is because of the following:

https://github.com/rails/propshaft/blob/69c285f293a4a845fbaa05625bbcb815e31ebded/lib/propshaft/output_path.rb#L55..L60

The regex ignores all files that have a `.digested` in them. 

Is this on purpose? or is this a bug?

tsrivishnu:
Looks like there a brief mention of this in https://github.com/rails/propshaft/issues/110

brenogazzola:
It's a bug. Are you using esbuild and allowing chunking?

tsrivishnu:
We allow chunking. Not with esbuild but with Webpack. I suppose the bundler doesn’t matter. Are there any attempts at fixing it that I can look at? Or known workarounds? For us, the assets directory grows to 40GB in a month or two and resolving that requires clobbering and precompile with a downtime of couple of minutes — a heavy JavaScript app that is going through a lot of iterations basically.  I’m going to test a script tomorrow that would look at the `.manifest.json` and deletes all the files with `.digested` in their name and are not mentioned in the manifest. On 29. May 2023, at 20:10, Breno Gazzola ***@***.***> wrote:﻿
It's a bug. Are you using esbuild and allowing chunking?

brenogazzola:
I think the problem is caused by #100 which I'm considering reverting as it's causing multiple problems, I just need confirmation. Last question: Do all chunks have the same base name and only differ in the digest?

I'm asking because clean is actually working for me, keeping only the 3 most recent versions of each file

brenogazzola:
Actually, I just figured it out. Clean is not going to work because each file name is unique due to the digest, and therefore Propshaft cannot group them into "versions" to delete the old ones.

I'll have to think a bit about how to handle this (I would also welcome a PR)

tsrivishnu:
Examples of pre-digested files:

```
public/assets/Vue3Lottie-1ea0f3b6ddc59bef7019.digested.js
public/assets/Vue3Lottie-23af1d5aa7baee8b1ca8.digested.js
public/assets/Vue3Lottie-26a7aeceee4e261a29b5.digested.js
[...]

PR #231, opened 2025-02-04
Cleanup pre-digested versions

schmijos:
This makes Propshaft responsible for `-[digest].digested` files in the output path. Propshaft copies them from the load path into the output path. So it should also clean them up (addressing #140).

I refrained from making `-[digest].digested` files into a `Propshaft::Asset`, because this proposal has less moving parts and @dhh doesn't want to invest in non-container setups were cleanup is needed at all.

### DIFF
--- a/lib/propshaft/output_path.rb
+++ b/lib/propshaft/output_path.rb
@@ -25,11 +25,13 @@ class Propshaft::OutputPath
   def files
     Hash.new.tap do |files|
       all_files_from_tree(path).each do |file|
-        digested_path = file.relative_path_from(path)
-        logical_path, digest = Propshaft::Asset.extract_path_and_digest(digested_path.to_s)
+        digested_path = file.relative_path_from(path).to_s
 
-        files[digested_path.to_s] = {
-          logical_path: logical_path.to_s,
+        digest = digested_path[/-([0-9a-zA-Z]{7,128})\.(digested\.)?([^.]|.map)+\z/, 1]
+        logical_path = digest ? digested_path.sub("-#{digest}", "") : digested_path
+
+        files[digested_path] = {
+          logical_path: logical_path,
           digest: digest,
           mtime: File.mtime(file)
         }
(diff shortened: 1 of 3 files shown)


brenogazzola:
I'm reading the code but I don't see anything that let's it handle the problem with esbuild where all chunks have the [same name](https://esbuild.github.io/api/#chunk-names).

For example, I have an empty `public/assets` folder, and after running `assets:precompile` I get this:
```
application-d70br74r.js
chunk-4hs4lcvp.js
chunk-iz6s2ho9.js
chunk-ftjr8syz.js
```

All the chunks above are fresh and required. Then after some changes in my JS file I run `assets:precompile` again and get this:
```

application-d70br74r.js
chunk-4hs4lcvp.js
chunk-iz6s2ho9.js
chunk-ftjr8syz.js
chunk-tc9xpclk.js
chunk-cy696x68.ks
```

Now, which are two stale chunks?

schmijos:
Ok, I see: there is no logical name (at least not without hash). Predigested files map 1-on-1 in the manifest. So we should not assume anything (the hash could even be the logical name in edge cases).

An idea to resolve this:
1. The version cannot be derived from the name but we can still cleanup predigested outputs by age (`1.hour` per default).
2. Because this could lead to a race condition where the previous version is still accessed but already gone (that's why `count is `2` by default, I guess), we would group predigested outputs by `:mtime` ranges.

Should I try this?

brenogazzola:
I just did a quick check, and if I run `rails assets:precompile`, esbuild does not update the timestamp of files it generates. So I'm don't think your one hour limit would work.

1 - Deploy at 3 AM, 5 chunks files with identical names and different hashes are generated
2 - Deploy at 5 AM, no change in the chunks, but they are now 2 hours out.
3 - The 1 hour limit means all of them are considered too old (except maybe 2) and will be removed.

schmijos:
Ok. I think this to be a dead end. Thank you for your review. I learned some things.

PR #235, opened 2025-02-19
Update UPGRADING.md

brendon:
When using multiple entrypoints, layouts need to be updated to refer to those files correctly.

Please correct me if I'm wrong, but this was my experience.


brendon:
I was wrong 🫢

PR #239, opened 2025-06-12
Remove railties runtime dependency

zcei:
This library technically has no dependency on `railties` to function.

Only if the environment we're in defines `Railtie` (Ruby on Rails) we want to load Rails specific behavior, but at this point we can rely on the gem already being loaded.

Making this change enables us to make it actually work in a standalone Rack environment, without implicitly changing the behavior of other gems that try to load their own railties because of the dependency being loaded.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,7 +5,6 @@ PATH
       actionpack (>= 7.0.0)
       activesupport (>= 7.0.0)
       rack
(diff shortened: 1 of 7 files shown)


andheiberg (commented):
I can confirm we have a real use-case for using Propshaft outside of Rails and that this would make that possible.

rafaelfranca (approved):

PR #240, opened 2025-06-24
Update UPGRADING.md guide

TimeTravelerFromNow:
This PR mentions that as as part of Upgrading steps, you should also remove other gems that list sprockets as a dependency


theodorton (approved):

theodorton:
I think we have to assume people understand they'll need to look for alternative gems when removing or upgrade/check for compatibility with existing ones.

Thanks!

PR #241, opened 2025-07-04
Improve CI coverage

rafaelfranca:
* Test with Ruby 3.4
* Test with all supported versions of Rails


theodorton:
You can ignore the rails 8/ruby 3.1 combo like this:
https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#excluding-matrix-configurations

PR #249, opened 2025-08-19
Turn Propshaft::Server into a proper middleware

### LINKED ISSUE #55508: ActiveRecord::Middleware::DatabaseSelector sets cookies on asset requests
Not sure if this is a bug, but noticed that when using the automatic DB switching ([docs](https://guides.rubyonrails.org/active_record_multiple_databases.html#activating-automatic-role-switching)), with the default resolver `ActiveRecord::Middleware::DatabaseSelector::Resolver` and context `ActiveRecord::Middleware::DatabaseSelector::Resolver::Session`, it seems like requests for assets are getting a cookie sent back in the response.

Noticed this while investigating issues with system tests not rendering flash messages correctly, and it looks like it was happening because of race conditions between redirect requests and asset requests, causing it to override the cookie and remove the flash message. This might be a different issue on our side (that we are not properly waiting for asset requests to finish before submitting a form or something like that), but it made me wonder if it's correct for asset requests to respond with a cookie.

I don't know too much about the session and cookie management, but from what I could debug, it seems like this is happening because the `ActiveRecord::Middleware::DatabaseSelector::Resolver::Session` middleware reads `request.session[:last_write]`, so it's like it makes it load the session, and maybe this makes `ActionDispatch::Session::CookieStore` send the cookie back in the response?

### Steps to reproduce
Created [this repo](https://github.com/EmilioCristalli/test_db_selector) to reproduce the issue. Has a few commits with the steps to create the app, but it's a normal Rails app with scaffold for one model (Post). 

The last commit in main has the automatic db switching enabled, but moving one commit back (to `9232d6768d`) should disable it to see the difference.

Can open the root page (which loads posts#index) and observe any asset requests (like http://localhost:3000/assets/application-bfcdf840.js) in the developer tools to see the `Set-Cookie` header in the response.

### Expected behavior
The asset requests shouldn't have a `set-cookie` header in the response.

(not 100% sure this is the expected behavior, but it looks like it's what happens when automatic DB switching is not enabled)

### Actual behavior
Asset requests have a `set-cookie` header in the response.

### System configuration

**Rails version**: 8.0.2.1
**Ruby version**: 3.4.5

byroot:
Fix: https://github.com/rails/rails/issues/55508

Mounting Propshaft::Server in the Rails router isn't suitable because it end up executed after a long list of middlewares that may set cookies or do other transformation we don't want.

In production, propshaft assets are served by `ActionDispatch::Static` (or a reverse proxy), so we should try to be as close to that as possible.

### DIFF
--- a/lib/propshaft/server.rb
+++ b/lib/propshaft/server.rb
@@ -2,30 +2,39 @@ require "rack/utils"
 require "rack/version"
 
 class Propshaft::Server
-  def initialize(assembly)
+  def initialize(app, assembly)
+    @app = app
     @assembly = assembly
   end
 
   def call(env)
     execute_cache_sweeper_if_updated
-    path, digest = extract_path_and_digest(env)
 
-    if (asset = @assembly.load_path.find(path)) && asset.fresh?(digest)
-      compiled_content = asset.compiled_content
+    path = env["PATH_INFO"]
+    method = env["REQUEST_METHOD"]
 
-      [
-        200,
-        {
-          Rack::CONTENT_LENGTH  => compiled_content.length.to_s,
-          Rack::CONTENT_TYPE    => asset.content_type.to_s,
-          VARY                  => "Accept-Encoding",
-          Rack::ETAG            => "\"#{asset.digest}\"",
-          Rack::CACHE_CONTROL   => "public, max-age=31536000, immutable"
-        },
-        [ compiled_content ]
-      ]
+    if (method == "GET" || method == "HEAD") && path.start_with?(@assembly.prefix)
+      path, digest = extract_path_and_digest(path)
+
+      if (asset = @assembly.load_path.find(path)) && asset.fresh?(digest)
+        compiled_content = asset.compiled_content
+
+        [
+          200,
+          {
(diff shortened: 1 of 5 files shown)


lib/propshaft/server.rb
```diff
@@ -2,30 +2,39 @@
...
-          Rack::ETAG            => "\"#{asset.digest}\"",
-          Rack::CACHE_CONTROL   => "public, max-age=31536000, immutable"
-        },
-        [ compiled_content ]
-      ]
+
+    path = env["PATH_INFO"]
+    method = env["REQUEST_METHOD"]
+
+    if (method == "GET" || method == "HEAD") && path.start_with?(@assembly.prefix)
```
skipkayhil, line 16: `HEAD` requests should return an empty body

https://github.com/rack/rack/blob/ee7ac5a1db5bc5c65e4b83342b8f4df88ef3c075/lib/rack/lint.rb#L819

byroot: `Rack::Head` would take care of that, but yes, good point.

PR #264, opened 2026-04-14
Add charset=utf-8 to Content-Type for CSS and HTML assets

flavorjones:
Note: this branch was based on https://github.com/rails/propshaft/pull/265 to get CI green.

### Motivation / Background

CSS files served by `Propshaft::Server` have `Content-Type: text/css` with no `charset=utf-8`. When Chrome discovers a stylesheet via `<link>` tag during HTML parsing (rather than via `Link: rel=preload` header), it may inherit the document's encoding or fall back to Windows-1252 instead of defaulting to UTF-8. This causes non-ASCII characters in CSS (e.g., en-dash U+2013 in `content` properties) to render as mojibake.

The same issue affects HTML assets.

The CSS spec defaults to UTF-8, but Chrome has a [longstanding bug](https://issues.chromium.org/issues/41118088) where it doesn't reliably follow this default. Adding `charset=utf-8` to the Content-Type header is the highest-priority encoding signal in the spec's detection chain.

Companion Rails PR: https://github.com/rails/rails/pull/57188

### Detail

`Propshaft::Server#call` now appends `; charset=utf-8` to the Content-Type header for `text/css` and `text/html` assets. Only these two MIME types are affected — `text/javascript` is excluded per RFC 9239, `text/xml` could break encoding declarations, and `text/plain` has no demonstrated browser bug.

### Additional information

`ActionDispatch::Response` already adds `charset=utf-8` unconditionally to all controller responses, but `Propshaft::Server` builds its own Rack response and bypasses that path. This fix closes that gap for CSS and HTML assets in development.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,12 @@ jobs:
           - "7.1"
           - "7.2"
           - "8.0"
+          - "8.1"
         exclude:
           - ruby-version: "3.1"
             rails-version: "8.0"
+          - ruby-version: "3.1"
+            rails-version: "8.1"
       fail-fast: false
     env:
       BUNDLE_GEMFILE: gemfiles/Gemfile.rails-${{ matrix.rails-version }}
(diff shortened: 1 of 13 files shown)


flavorjones:
Failing tests would be fixed by pinning minitest, see https://github.com/rails/propshaft/pull/265

PR #268, opened 2026-07-03
Don't use Mime::EXTENSION_LOOKUP directly

etiennebarrie:
See https://github.com/rails/rails/pull/57826, these constants are now deprecated in favor of public APIs.

### DIFF
--- a/lib/propshaft/load_path.rb
+++ b/lib/propshaft/load_path.rb
@@ -33,7 +33,7 @@ class Propshaft::LoadPath
 
   def asset_paths_by_type(content_type)
     (@cached_asset_paths_by_type ||= Hash.new)[content_type] ||=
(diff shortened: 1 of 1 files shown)


lib/propshaft/load_path.rb
```diff
@@ -52,7 +52,7 @@ def manifest
   # and test to ensure the map caches are reset when javascript files are changed.
   def cache_sweeper
     @cache_sweeper ||= begin
-      exts_to_watch  = Mime::EXTENSION_LOOKUP.map(&:first)
+      exts_to_watch  = Mime.respond_to?(:extensions) ? Mime.extensions : Mime::EXTENSION_LOOKUP.map(&:first)
```
rafaelfranca, line 55: We should add a TODO to remove this check when we drop support to the versions of Rails that don't have this method

hmcguire-shopify: https://github.com/rails/propshaft/pull/276

PR #272, opened 2026-07-29
Look up :all stylesheets through the resolver

hmcguire-shopify:
Currently, `all_stylesheets_paths` asks `LoadPath` for every `text/css` asset, which requires walking every configured asset path and building an `Asset` for each file found. That work is unnecessary in production, where `Resolver::Static` otherwise resolves digested paths straight out of the precompiled `.manifest.json`.

This commit moves the lookup to the resolver, so `Resolver::Static` can use the manifest while `Resolver::Dynamic` delegates to to current `LoadPath` logic.

`:app` still needs the load path (for now), since it globs source paths that the manifest doesn't record.

Removing usage of the `LoadPath` in production helps get us closer to using `Propshaft::Assembly` inside a Ractor.

### DIFF
--- a/lib/propshaft/helper.rb
+++ b/lib/propshaft/helper.rb
@@ -29,7 +29,7 @@ module Propshaft
   # == Bulk Stylesheet Inclusion
   #
   # The stylesheet_link_tag helper supports special symbols for bulk inclusion:
-  # - :all - includes all CSS files found in the load path
+  # - :all - includes all CSS files known to the asset resolver
   # - :app - includes only CSS files from app/assets/**/*.css
   #
   #   <%= stylesheet_link_tag :all %>  # All stylesheets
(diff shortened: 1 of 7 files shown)


lib/propshaft/helper.rb
```diff
@@ -105,9 +105,9 @@ def javascript_include_tag(*sources)
       _build_asset_tags(sources, options, :javascript) { |source, opts| super(source, opts) }
     end
 
-    # Returns a sorted and unique array of logical paths for all stylesheets in the load path.
+    # Returns a sorted and unique array of logical paths for all stylesheets known to the resolver.
     def all_stylesheets_paths
```
Edouard-chin, line 109: It may be interesting to add a test for this method and ensure the returned values is the expected on no matter what resolved is being used

rafaelfranca (approved):
lol, and here we go implementing sprockets again.

Yes, this makes sense. It should be always like this.

